PR #1, opened 2020-03-23
necklace

kimsappi:
There are some hardcoded shortcuts for cases where strlen is different between s1 and s2.

I solve the main problem by rotating s2 by trying to copy the end of the string beginning from index i and concatenating the beginning of the string. If the strings are considered matching, this will cause them to become identical after looping >= strlen times.

### DIFF
--- a/chall00/ksappi.c
+++ b/chall00/ksappi.c
@@ -1,0 +1,21 @@
+#include <string.h>
+
+int hv_necklace(char *str1, char *str2)
+{
+    if (strlen(str1) != strlen(str2))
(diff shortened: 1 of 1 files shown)


jgengo:
If you do use the size of a string multiple times, it would be better to store it in a variable instead of using many times strlen.

You could have avoided using malloc (strdup) for this challenge if you would have use strncmp instead of strcmp.  If you allocate memory you should also make sure to free it.

Super clean code and pushed in less than 15 minutes!   Good job! 👍

PR #2, opened 2020-03-23
chall00

vkuikka:
The function takes the length of first string and rotates the other string that amount of times and tries to match them on each rotation.

### DIFF
--- a/chall00/vkuikka.c
+++ b/chall00/vkuikka.c
@@ -1,5 +1,3 @@
-#include <unistd.h>
-
 int		ft_strequ(char *st1, char *st2)
 {
 	while (*st1 && *st2 && *st1 == *st2)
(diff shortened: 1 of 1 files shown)


vkuikka:
Didn't add a description so i'm adding it here. The function takes the length of first string and rotates the other string that amount of times and tries to match them on each rotation.

plantran:
Hey!
Nice work, I appreciate that you used the `libc` functions as much as possible.

Here are some comments and suggestions regarding your code:

- Maybe it would be better to assign your `len` variable before doing your first check so that you don't call strlen a second time.
- You could have used `strdup` directly (and still do your checks).
- One thing that you could have done is assigning the result of your `strequ` in your while so that you don't have to call it again after:
```
while (!(b = ft_strequ(str1, str2)) && i++ < len)
		ft_rotate(str2);
	if (b)
		return (1);
```

In my opinion it still makes the code readable!
Good job!

PR #3, opened 2020-03-23
solution

wehard:
First find check lengths are equal.
Next find index of first char in s1 in s2.
Then loop and check equality using mod to wrap index.

### DIFF
--- a/chall00/wkorande.c
+++ b/chall00/wkorande.c
@@ -1,0 +1,38 @@
+/* ************************************************************************** */
+/*                                                                            */
+/*                                                        :::      ::::::::   */
+/*   wkorande.c                                         :+:      :+:    :+:   */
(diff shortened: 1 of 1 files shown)


jiricodes:
It doesn't work on words that have multiple same letters e.g.:
`hv_necklace("solosa", "losaso")`

wehard:
> It doesn't work on words that have multiple same letters e.g.:
> `hv_necklace("solosa", "losaso")`

Yep noticed that too :face_with_head_bandage: It did feel too simple

jgengo:
You could have avoid storing strlen(s1) while you only use it one time.

You include stdio.h but you are not using it. You should avoid including unused libs.

I think this would segfault, let’s wait for the auto-evaluation trace.

Good job and nice try still ! 👍

PR #4, opened 2020-03-23
Necklace (jnovotny)

jiricodes:
Simple one by one letter comparison, looping around the s1 from different starting points.

### DIFF
--- a/chall00/jnovotny.c
+++ b/chall00/jnovotny.c
@@ -1,0 +1,52 @@
+/* ************************************************************************** */
+/*                                                                            */
+/*                                                        :::      ::::::::   */
+/*   necklace.c                                         :+:      :+:    :+:   */
+/*                                                    +:+ +:+         +:+     */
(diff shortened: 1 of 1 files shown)


plantran:
Hi,
Nice work!

However, here are some suggestions regarding your code:

- You should prefer using the libc `strlen` function or if you want to code yours, prefer prototyping it as a `size_t` because `strlen` can only return positive values or 0 so better to use the proper range (int range: [-2147483648, 2147483647] ; size_t range [0, 4294967295])
- You could change this part :
```
l = ft_strlen(s1);
	if (l != ft_strlen(s2))
		return (0);
	else if (l == 0 && ft_strlen(s2) == 0)
		return (1);
```

with 

```
l = ft_strlen(s1);
if (l != ft_strlen(s2))
  return (0);
else if (l == 0)
  return (1)
```
That way you don't call `strlen(s2)` twice, as you already know they have the same length. However it's good you stored `strlen(s1)` in the first place 👍  

Otherwise I like this simple yet effective implementation, good job!

PR #5, opened 2020-03-23
challenge00

ptuukkan:
This applies slide function and compares the strings until all combinations have tested.


jgengo:
You include stdio.h but you are not using it. You should avoid including unused libs.

You could have avoided using malloc for this challenge if you would have use strncmp instead of strcmp. You also should avoid using malloc it was not necessary.

When you use malloc to make sure to free the memory allocated.

You used the libc strlen, strcmp you could also have used the strdup one you would have saved some time.
 Code is clear and clean!  

Good job! 👍

PR #6, opened 2020-03-23
Add my submission sadawi.c

salahadawi:
I solved this exercise by rotating s2 to every possible orientation, then checking each of the orientations to see if they match with s1. If any of them match, the function returns 1, else it returns 0.

I think a faster implementation would be to use some kind of linked list maybe? If the strings are too large it could take quite a while to rotate through all orientations.

### DIFF
--- a/chall00/sadawi.c
+++ b/chall00/sadawi.c
@@ -1,0 +1,117 @@
+/* ************************************************************************** */
+/*                                                                            */
+/*                                                        :::      ::::::::   */
+/*   sadawi.c                                           :+:      :+:    :+:   */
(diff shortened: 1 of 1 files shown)


plantran:
Hi,
The code is clean and the PR description understandable.
Here are some feedback/suggestions for your code:

- You can use the libc `strlen()` function. If you want to use your own, then you should prefer prototyping it as `size_t` because: `strlen` can only return positive values or 0 so better to use the proper range (int range: [-2147483648, 2147483647] ; size_t range [0, 4294967295])
- You didn't necessarily need to manipulate memory to resolve this assignment.

But overall, good job, merging this pull request.

PR #7, opened 2020-03-23
Create challenge 00

lapaset:
In my solution i go through string one and every time i run into the first character of string two, i check if it's a match.


jgengo:
Code super clean and clear.

No extra functions used or a library, not using malloc.

From what I can see it works, let's wait for the auto evaluation trace now.

Sadly, I’ve nothing to say except a good job! 👍

PR #9, opened 2020-03-23
Add solution

hiiri:
Copy s1 to itself so the same string appears 2 times, then compare s2 to the new string. If the whole s2 string matches anywhere in the new concatenated string, it means it is continuously rearranged

### DIFF
--- a/chall00/alcohen.c
+++ b/chall00/alcohen.c
@@ -1,0 +1,50 @@
+#include <stdlib.h>
+
+int len(char *s)
+{
+    int i = 0;
+    while (s[i])
+        i++;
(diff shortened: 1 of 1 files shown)


jgengo:
You created a function called `len()`, if you want to make your own function try to follow the convention and name it `my_strlen()` for example.

You are using len(s1) multiple times so I would suggest storing it in a variable.

You also should avoid using malloc it was not necessary. If you use it to make sure to free the memory allocated.

Good and clear, from what I can see it works. Let’s wait for the auto-evaluation trace now.  Good job! 👍

PR #10, opened 2020-03-23
Finished!(add descritption)

ihwang:
Summary
1. make all possible family members of original string into double pointer.
2. compare each family member to our second parameter.
3. free the double pointer.


plantran:
Hi,
The code is clean and the Pull request description is informative.
This implementation is rather original, however, here are some feedback/suggestions regarding your code:

- You can use the libc `strlen()` function. If you decide to use your own, prefer prototyping it as a `size_t` because `strlen` can only return positive values or 0 so better to use the proper range (int range: [-2147483648, 2147483647] ; size_t range [0, 4294967295]).

- Although your method works, it requires generating all the different possible combinations, which can be very resourceful if the string is very long and it's not the most optimal way. Instead of making all the examples, you could have looped `len` times and compare your original string to the newly rotated one. If there's a match, then you return 1.

- Protect your mallocs, always!

- There were ways to resolve this assignment without allocating any memory.

However this is a good piece of code, I'm merging this pull request! 👍

ihwang:
Hi.

I appreciate your thorough code review!
I couldn’t expect you to review my code deeply like this.
All of these processes must have taken your time a lot, It’s much thankful to me.
And I will keep your suggestions in my mind.
Have a great day!

PR #11, opened 2020-03-23
my version

DianaMukaliyeva:
First I checked length of the words. If they are not equal, we return false, otherwise, if they are equal to 0, then our function return true. Then I move characters in first string and compare it with another string.

### DIFF
--- a/chall00/dmukaliy.c
+++ b/chall00/dmukaliy.c
@@ -1,0 +1,41 @@
+
+#include <string.h>
+
+char	*move_to_end(char *str, size_t len)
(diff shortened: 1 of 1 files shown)


jgengo:
Really clean code.

strdup is using malloc and for this challenge, I think you could avoid allocating memory for this challenge if you would have use strncmp instead of strcmp. 

I notice you stored strlen(s1) because it’s used multiple times while strlen(s2) isn’t stored because it’s only used one, that’s really good!

Good job! 👍

PR #12, opened 2020-03-23
Chall00 solution

tdawson-42:
Copy s1 twice so that it repeats itself, ie. "nicole" becomes "nicolenicole". s1 represents the same necklace as s2 if the two strings are equal lengths and the string containing s1 twice (s1s1) contains s2 within itself.


plantran:
Very nice solution!
The code is very clean and the pull request description is enough descriptive.
Good job, I'm merging this pull request! 👍

PR #13, opened 2020-03-23
asolopov solution

Artemso:


Artemso:
1) create temp string
2) shift x chars to the back
3) compare to s2

jgengo:
You created a function called `get_len()`, if you want to make your own function try to follow the convention and name it `my_strlen()` for example.
 You should use the libc function on speed coding contests, but it’s up to you if you do prefer coding them you can.

You also should avoid using malloc it was not necessary. But if you do I suggest you not to use implicit casting when using it. 
And also you should have include stdlib.h instead of stdio.h, always make sure you don’t include unused lib.

You should also propose a PR only when you are done and sure it works to looks more professional and try to describe a little bit more your code in the PR description

Good job though!  👍

Artemso:
Thanks for the feedback, I’ll do better next time. Thanks!

PR #14, opened 2020-03-23
rjaakonm

rasmusjaa:
Rotate the first string as many times as possible (length of the first string) to test all possible variations and check if one those matches the second string.

### DIFF
--- a/chall00/rjaakonm.c
+++ b/chall00/rjaakonm.c
@@ -1,0 +1,49 @@
+#include <stdio.h>
+#include <string.h>
+#include <stdlib.h>
+
+char	*rotone(char *s1)
+{
+	int len;
+	int i;
+	char x;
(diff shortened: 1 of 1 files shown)


plantran:
Hello,
The code is clean and the pull request description is understandable.
However, here are some feedback and suggestions regarding your code:

- Don't forget to remove all unnecessary includes after you're done testing (`stdio.h`).
- `strlen()` returns a `size_t`, you should use that type too, especially because in that case it will allow you to manipulate a wider range of numbers.
- You could have sent `len` to your `rotone` function so you don't have to call that function everytime.
- Don't forget to protect your mallocs 🐊
- There were ways to solve that assignment without manipulating the memory.

Anyway, the code is clean, I'm merging this Pull Request 👍

PR #15, opened 2020-03-23
chall00 v1.0.0

MikeyLHolm:
Few protects and checks at start to avoid useless looping.

Compare s2[j] to s1[0]. When match is found traverse both strings comparing them. If s1 reaches '\0' strings match. If mid traverse s2 ends, its index is reseted to 0 while s1 index stays. If s1 never reaches the end: return (0);

### DIFF
--- a/chall00/mlindhol.c
+++ b/chall00/mlindhol.c
@@ -1,0 +1,51 @@
+/* ************************************************************************** */
+/*                                                                            */
+/*                                                        :::      ::::::::   */
+/*   mlindhol.c                                         :+:      :+:    :+:   */
+/*                                                    +:+ +:+         +:+     */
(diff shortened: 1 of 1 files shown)


jgengo:
Quite clear code.

You could have avoid declaring and assigning strlen(s1) while you only use it one time.

You can also change 

```c
s1[i] != “\0”
```
by
```c
!s1[i]
```

to have this:

```c
while (s1[i] != '\0' && s2[j] != '\0' && s1[i] == s2[j])
```

looking like:

```c
while (!s1[i] && !s2[j] && s1[i] == s2[j])
```

I do like the fact you come up with a solution using `strchr()`

Good job! 👍

PR #16, opened 2020-03-23
Add chall 00

hakolao:
Compare characters throughout second string starting from any position and stop on first match. Or false otherwise.

No libs :)

### DIFF
--- a/chall00/ohakola.c
+++ b/chall00/ohakola.c
@@ -1,0 +1,52 @@
+static int	ft_strlen(char *str)
+{
+	int	i;
+
+	i = 0;
+	while (str[i])
+		i++;
+	return (i);
+}
+
+static int cmp(char *s1, char *s2, int len, int pos2)
+{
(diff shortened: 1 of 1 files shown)


plantran:
Hello!
The code is clean and the PR description is informative enough.

Here are some feedback or suggestions regarding your code:
- When creating your own `strlen()` function, prefer prototyping it as a `size_t` like the libc one because `strlen` can only return positive values or 0 so better to use the proper range (int range: [-2147483648, 2147483647] ; size_t range [0, 4294967295]).
- You don't have to check if the length of both strings is the same twice:
```
if (len != len2)
  return (0);
if (len == len2 && len == 0)
  return (1);

to

if (len != len2)
  return (0);
if (len == 0)
  return (1);
```
In my own personal opinion, it still makes the code readable and understandable. 

Other than that, the solution is simple yet very effective, good job! 👍 
I'm merging this PR.

PR #17, opened 2020-03-23
krusthol's solution for hv_necklace

krusthol:
This solution relies firstly upon checking any length difference between both strings. If the difference is not equal, return 0. If the lengths are equal, keep rotating through all the possibilities by rotating the second string and checking the strings with string compare, returning 1 if match found, or eventually returning 0 if no matches are found. Solution uses <stdlib.h> for malloc and <string.h> for strcmp.

### DIFF
--- a/chall00/krushtol.c
+++ b/chall00/krushtol.c
@@ -1,0 +1,64 @@
+#include <stdlib.h>
+#include <string.h>
+
+int		compare_len(const char *s1, const char *s2)
+{
+	int s1_len;
+	int s2_len;
+
+	s1_len = 0;
+	s2_len = 0;
+	while (*s1++)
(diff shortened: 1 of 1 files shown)


jgengo:
You could have avoid using malloc and free for this challenge if you would have use strncmp instead of strcmp. You also should avoid using malloc it was not necessary.

Also I would suggest change

```c
	neck_len = compare_len(s1, s2);
	if (neck_len < 0)
		return (0);
```


by 

```c
	if (strlen(s1) != strlen(s2))
		return (0);
```

it makes it cleaner and avoid you to code another function

However, you came with the right algorithm to solve this problem and your code is clean and clear though.

Good job! 👍

PR #18, opened 2020-03-23
hv_necklace

HandsomeTom:


plantran:
Hello!
Unfortunately, you didn't provide any description with your Pull Request explaining how you solved this assignment so I won't be merging it.

Now, regarding your code, here are some feedback and suggestions:
- You can use the libc `strlen()` function. If you prefer using your own, then you should prefer prototyping it as a `size_t` because `strlen` can only return positive values or 0 so better to use the proper range (int range: [-2147483648, 2147483647] ; size_t range [0, 4294967295]).
- You should assign your `ft_strlen(s1)` to a variable earlier in your code to prevent having to call it several times.
- I don't think you need that first part here
```
if (ft_strlen(s1) == 1)
{
  if (*s1 == *s2)
    return (1);
  else
    return (0);
}
```
 as it is not a special case and it will be checked later in your code anyway.
- Your solution generates a `bus error` as you are manipulating `str` directly in your `swap_char` function, I suggest you take a deeper look at that :).

Other than that, good job still 👍

PR #19, opened 2020-03-23
My solution

tuommii:
Moving characters from start to end until all possible solutions has been compared


jgengo:
You could have avoided using malloc and free for this challenge if you would have use strncmp instead of strcmp. You also should avoid using malloc it was not necessary.

You include stdio.h but you are not using it. You should avoid including unused lib.
 Good and clean code, you also used size_t type, I appreciate.

Good job! 👍

PR #20, opened 2020-03-23
My solution to chall00

MuSuareZ:
My function does exactly the same logic as the necklace letters described, where "first_to_next" function rotates the second string letters to their next position until the length of the first string is matched. If our first string is equal to our second string, we return (1), otherwise we can say they don't match the necklace rules and instead we return (0).


jgengo:

I like the fact you used strequ to solve this challenge and declare your function with static.

Good job! 👍

PR #21, opened 2020-03-23
Add hhuhtane.c which contains my version of hv_necklace funtion

huhtasheikki:
1. Check that both lines are equally long.
2. Try to find String s1's first character from s2
3. When found matching starting characters, start comparing strings. When reached to end of s2 continue from the start. If any characters doesn't match, the function returns 0.

### DIFF
--- a/chall00/hhuhtane.c
+++ b/chall00/hhuhtane.c
@@ -1,0 +1,50 @@
+/* ************************************************************************** */
+/*                                                                            */
+/*                                                        :::      ::::::::   */
+/*   hhuhtane.c                                         :+:      :+:    :+:   */
+/*                                                    +:+ +:+         +:+     */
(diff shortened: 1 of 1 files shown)


jgengo:

You can use strlen of libc. However, if you planned to use your libft functions (which is really nice) for ft_strlen I would recommand size_t return type instead of int return type:

I’m not sure it will pass the tests, I’m a bit confused but let’s see what the auto evaluation trace says.

but if I try this, it seems you return 0 while it should returns 1 

```c
hv_necklace("rabcr", "abcrr”))
```



  I would suggest another condition inside your ft_strlen 

```c
while (str && str[i])
  i++;
```

by

```c
while (*str++)
  i++;
```

Good job! 👍

PR #22, opened 2020-03-23
My solution for the hv_necklace -kmoilane

kmoilane:
My solution for the hv_necklace coding challenge. I used stdlib functions, which I also have in my libft but I thought I'd use the originals for once. 26 lines, but could also be fit in 25.

### DIFF
--- a/chall00/kmoilane.c
+++ b/chall00/kmoilane.c
@@ -1,0 +1,32 @@
+#include <stdlib.h>
+#include <string.h>
+
+int	hv_necklace(char *s1, char *s2)
+{
+	char	swap;
+	char	*temp;
+	int	i;
+	int	rotations;
+	int	len;
+
+	len = strlen(s2);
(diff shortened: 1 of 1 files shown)


jgengo:
When you use `strlen` multiple times you should store it in a variable, I can see you stored strlen(s2) but just a few lines under you use strlen(s2) again instead of using your variable `len`

You used malloc and I don’t think it was necessary for this challenge, and you do not free the memory allocated.

The code could have been a little bit cleaner, but from what I can see it looks to work, let’s wait for the auto-evaluation trace for a confirmation.

Good job! 👍

kmoilane:
Thanks for the feedback! Yeah it was a bit rushed and I merged couple of
ideas quickly together. Next time I'll focus more on cleaner code.

ti 24. maalisk. 2020 klo 13.46 Jordane GENGO <notifications@github.com>
kirjoitti:

PR #23, opened 2020-03-23
challenge 1. hv_necklace by sluhtala

sluhtala:
I use something that is basically an strcmp but with an offset value. I call it ft_strcmp_offset.
I loop that function while changing the offset value until the value equals the length of the string.


jgengo:


Using `ft_strcmp` with an offset was imho a perfect choice, the function is named : `strncmp` 

Good and clean code, really nice

Good job! 👍

PR #24, opened 2020-03-23
Solution for chall00

AleXwern:
No extra libraries used and that was the goal.
Since I cannot modify read-only strings, I created a modified ft_strncmp that uses an offset index that loops through the string s1 and this index is modified with +1 during every loop and the entire thing loops N times where N is the length of the array.
If the offset index reaches end of string s1 but not the s2 it simply rolls to offset+index = 0, otherwise both strings are considered complete.


jgengo:
There is a problem with your header. Let’s skip the headers for those remote-challenges but appreciate the effort to use it from home.

Clean code, good PR description and usage of strncmp with index was a perfect choice imho.

Good job! 👍

PR #25, opened 2020-03-23
elindber

einoob:


jgengo:
I notice you stored strlen(s1) because it’s used it multiple times while strlen(s2) isn’t stored because it’s only used one, that’s really good!
 Good idea to have use push_back and strcmp it makes your solution really clean.

However, you used malloc and I don’t think it was necessary for this challenge, and you do not free the memory allocated.

Unfortunately, you didn’t write any PR description so I've to close this PR.

Good job though! 👍

PR #26, opened 2020-03-23
ykalashn

ykalashn:
1. Comparing length of the strings and moving character by character in string 1 trying to find similarity in string 2. 
2. If the similarity found in string 2, move to the next character in string 2. 
3. It returns `1` if the string 2 reached its end.


jgengo:
Your header email seems to say you are from 42 Paris :) Let’s skip the headers for those remote-challenges but appreciate the effort to use it from home.

Quite clear code, tiny and efficient, sadly I think it won’t pass the empty string test, but let’s wait the auto evaluation trace.

Sadly, I don’t have much more to say about your code it’s pretty good.

Good job! 👍

ykalashn:
Thank you for the feedback! I really appreciate that you checked my code. 
Yeah, I did not notice the last example of what the function does, so it would return `0` in case with empty strings. Also, I agree we should skip the headers 🤗

Can't wait for the next challenge! 🚀

PR #27, opened 2020-03-23
hopham's solution

shilena91:
This is my solution. I used some functions from built-in library but they're all from libft :). Hope its ok. I also don't check norm :).


jgengo:
You were asked to send a function you submitted a program.

You should avoid using malloc it was not necessary and because of strdup and all the malloc used you have to use free a lot which make the code less easy to read.

Sadly, from what I’m able to see it looks like it would have worked without the main function.

Unfortunately, your PR description isn’t talking about your code so much.

Good job though! 👍

PR #28, opened 2020-03-23
My solution for chall00

timohive:
There are better ways to do this but this was my first idea so I went with that.


jgengo:
(I do like your hive logo profile picture)

Unfortunately, your PR description isn’t talking about your code, but I let it pass because you have added really good comments into your code. 
You also should avoid using malloc it was not necessary.

I notice you stored strlen(s1) because it’s used it multiple times while strlen(s2) isn’t stored because it’s only used one, that’s really good!

Let’s wait for the auto evaluation, but from what I can see it works pretty fine.  Clean code and good usage of strcmp

Good job!

PR #29, opened 2020-03-23
Added my own hv_necklace function

VictoriaElisabet:
I started by solving the problem on paper, and I figured that string s2 should be seen like a string with no end just like a necklace (a string on a loop) meaning that I will jump over the null terminator if needed.
I needed two while loops, one to loop through s2 string to find out if it contains the first letter of string s1. 
In case I find a match I will go into the other while loop where I will start checking if the next letter in s2 is the same as in string s1. 
If I looped through the whole s1 string I will return 1, otherwise I will return 0.
I also check if both strings are the same lenght.
Not the most beautiful solution in the world :)

### DIFF
--- a/chall00/vgrankul.c
+++ b/chall00/vgrankul.c
@@ -1,0 +1,71 @@
+/* ************************************************************************** */
+/*                                                                            */
+/*                                                        :::      ::::::::   */
+/*   hw_necklace.c                                      :+:      :+:    :+:   */
+/*                                                    +:+ +:+         +:+     */
+/*   By: vgrankul <marvin@42.fr>                    +#+  +:+       +#+        */
+/*                                                +#+#+#+#+#+   +#+           */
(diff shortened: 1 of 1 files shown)


jgengo:
Good PR description and clean code!

You can use strlen of libc. However, if you planned to use your libft functions (which is really nice) for ft_strlen I would recommand size_t return type instead of int return type:

check_match function is really clear and clean. I just don’t really understand why you did not declare and assign `i`, `j` and `k` inside this function instead of passing it into your function arguments.

Let’s wait for the auto evaluation, but from what I can see it works.

Good job! 👍

VictoriaElisabet:
Thank you!
I think I was so stuck in the idea to not exceed 25 lines in a function and that is why I passed it as arguments instead, would probably have been better to declare them in the function :)

PR #30, opened 2020-03-23
tkuumola solution chall00

T7Q:
1. compared length of strings, continued if they matched
2. created a string "temp" to store initial s1 and its variations.
3. in the while loop compared temp and s2 for the length of s1:
      - if match found: returned 1
      - if not: updated temp with strdup based on position of i and concatenated the letters before i with strlcat. 
4. if conditions were not met returned 0

### DIFF
--- a/chall00/tkuumola.c
+++ b/chall00/tkuumola.c
@@ -1,4 +1,5 @@
 #include <string.h>
+#include <stdlib.h>
 
 int hv_necklace(char *s1, char *s2)
 {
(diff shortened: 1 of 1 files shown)


jgengo:
Good PR description, clean code and seems to handle all the cases, let’s wait for the trace.

  I would suggest to store strlen(s1) in a variable like it looks like you use it multiple times. 
strdup is using malloc and for this challenge, I think you could have avoided allocating memory.

Really nice solution!   Good job! 👍

T7Q:
thank you very much for giving feedback!

PR #31, opened 2020-03-23
bling

Rakutaa:


Rakutaa:
maybe.. ?

jgengo:
Good and clean code  .

I would also recommend using libc functions especially in speed coding contests, but it’s up to you to use your libft functions.

Unfortunately, your PR description isn’t enough describing your code.

  Good job though! 👍

PR #32, opened 2020-03-23
solutions of Neckles challenge

mhjony:


jgengo:
You did include stdio.h and unistd.h, be careful not to import unused libs

You should also avoid static allocation with like `char str_cpy[4096];`

You should also try to split your codes in multiple small functions, for example implementing a strlen inside hv_necklace makes the code a little bit less clean. I would suggest using strlen or add your ft_strlen function.

Otherwise, it’s clean and I think it handle the cases of the tests. Let’s wait for the trace.  

Unfortunately, you did not add any description to your PR.

Good job though! 👍

jgengo:
Unfortunately, you did not add any description to your PR.

PR #33, opened 2020-03-23
spentti solution for chall0

saraarmiida:
I check if strings are the same length and if so, i rotate the characters in the second string to check if any of the rotations match the first string.


jgengo:

I would recommend using libc functions especially in speed coding contest, but it’s up to you to use your libft functions.

You also should avoid using malloc it was not necessary.

Clean code, and seems to handle all the cases, let’s wait for the trace.  

Good job! 👍

PR #34, opened 2020-03-23
rpehkone

vossr:


jgengo:
You did include stdio.h, be careful not to import unused lib

I notice you stored strlen(s1) because it’s used multiple times while strlen(s2) isn’t stored because it’s only used one, that’s really good!

You also should avoid using malloc it was not necessary. If you use it also you should remember to free the memories allocated.

Good job! 👍

PR #35, opened 2020-03-23
jhakala

juhakala:
1. Find the first letter of s1 from s2
2. Check with while that s1 can be made with s2 from that point, if not then go to next letter that matched to s2

### DIFF
--- a/chall00/jhakala.c
+++ b/chall00/jhakala.c
@@ -1,0 +1,44 @@
+int ft_strlen(char *str)
+{
+  int i;
+  
+  i = 0;
+  while (str[i])
+    i++;
+  return (i);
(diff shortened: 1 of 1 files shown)


jgengo:
You can use strlen of libc. However, if you planned to use your libft functions (which is really nice) for ft_strlen I would recommend size_t return type instead of int return type

You should avoid variable assignation into condition expression as you did on line 19. You don’t have to respect the Hive norm so I would recommend taking more lines to make sure your code is easier to read or even split it into small functions.

I think you forget something in the condition line 17 or 19, none of them check empty string.

 Let’s see the trace later.

  Good job! 👍

PR #36, opened 2020-03-23
Chall00 - the Necklace - hkoskela

glacela:
At first I check some specific cases like if both strings are empty, one of them is, different length or already the same.

Comparing to the second string I allocate a new string where I move the characters one backwards and then the first char to the last position and keep comparing until I've ran out of rounds or found a match with the first string.

I use two functions from my Libft project ft_strcmp and ft_strlen with int.

### DIFF
--- a/chall00/hkoskela.c
+++ b/chall00/hkoskela.c
@@ -44,7 +44,7 @@ int     hv_necklace(char *s1, char *s2)
     
     i = 0;
     r = ft_strlen(s1);
-    if (!s1 && !s2 || s1 == s2)
+    if (!s1 && !s2 || ft_strcmp(s1, s2) == 0)
       return (1);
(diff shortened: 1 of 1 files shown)


jgengo:
You can use strlen of libc. However, if you planned to use your libft functions (which is really nice) for ft_strlen I would recommend size_t return type instead of int return type

You also should avoid using malloc it was not necessary. If you use it also I would suggest to explicitly cast your malloc on line 29 and also free the memories allocated


I notice you stored ft_strlen(s1) because it’s used 2 times while ft_strlen(s2) isn’t stored because it’s only used one, that’s really good! 
 About the condition [line #47](https://github.com/hivehelsinki/remote-challs/pull/36/files#diff-12f173f8933d53b373f7c62247a4ba13R47) I think it will probably raise a warning and you should place parentheses around the `&&` expression.

Good job! 👍

PR #37, opened 2020-03-23
Necklace

mashalink:
Check if s1 or s2 = NULL or if length of s1 != length of s2 - return 0.

Check if lengths of s1 and s2 = 0 - return 1.

Use loop to compare one by one chars from s1 with the first char from s2.
If conditions met, use nested loop to compare one by one next chars from s1 and s2 and till s1 exists.
If s1 != '\0' - break the loop.
Continue using nested loop to compare s1 and s2 starting from the first char of s1 and till s2 exists.
If after comparing s2 = '\0' - return 1.

If no return 1 during the loop - return 0.

### DIFF
--- a/chall00/mlink.c
+++ b/chall00/mlink.c
@@ -1,0 +1,42 @@
+int ft_strlen(char *s)
+{
+    int	i;
+
+	i = 0;
+	while (s[i] != '\0')
+		i++;
+	return (i);
+}
+
+int hv_necklace(char *s1, char *s2)
+{
+    int len;
+    int i;
+    int j;
+    int x;
+
+    x = -1;
+    if (!s1 || !s2 || (ft_strlen(s1) != ft_strlen(s2)))
+        return (0);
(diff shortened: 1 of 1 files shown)


mashalink:
Check if s1 or s2 = NULL or if length of s1 != length of s2 - return 0.

Check if lengths of s1 and s2 = 0 - return 1.

Use loop to compare one by one chars from s1 with the first char from s2.
If conditions met, use nested loop to compare one by one next chars from s1 and s2 and till s1 exists.
If s1 != '\0' - break the loop.
Continue using nested loop to compare s1 and s2 starting from the first char of s1 and till s2 exists.
If after comparing s2 = '\0' - return 1.

If no return 1 during the loop - return 0.

jgengo:

Plus, you use ft_strlen(s1) and ft_strlen(s2) multiple times, I would recommend storing in a variable the return value of each so it would avoid recounting the size of s1 and s2 every time you need it.

The code could be a little bit cleaner with few modifications but looks good though.

Good job! 👍

PR #39, opened 2020-03-24
my code

Biffenx:
Simple code (But hard enough for me..).
- I check the s1 and s2 strings length} if they are not the same I return 0.
- i do two while loops checking for matching letters, If s2 hits the end of string I change the index to 0. 
- if the loop for s1 comes to the end and matching all the way then return 1.

 A big hand to my fellow peers jnovotny and mlindhol for helping me with the PR and jnovotny for also spotting errors in the code.

### DIFF
--- a/chall00/jwilen.c
+++ b/chall00/jwilen.c
@@ -6,12 +6,9 @@
 /*   By: jochumwilen <jochumwilen@student.le-101.fr>+#+  +:+       +#+        */
 /*                                                +#+#+#+#+#+   +#+           */
 /*   Created: 2020/03/23 21:59:45 by jochumwilen       #+#    #+#             */
-/*   Updated: 2020/03/23 21:59:45 by jochumwilen      ###   ########lyon.fr   */
+/*   Updated: 2020/03/24 10:05:56 by jochumwilen      ###   ########lyon.fr   */
 /*                                                                            */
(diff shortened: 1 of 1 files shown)


jgengo:
Your header seems to say you are from 42 Lyon :) Let’s skip the headers for those remote-challenges but appreciate the effort to use it from home.

You use ft_strlen only one time for each string, so I would recommend avoiding declaring len_s1 and len_s2 and directly use it in your condition 

```c
int len_s1;
int len_s2;
len_s1 = ft_strlen(s1);
len_s2 = ft_strlen(s2);
if (len_s1 != len_s2)
  return (0);
```

  by 

```c
if (ft_strlen(s1) != ft_strlen(s2))
  return (0);
```

 Good usage of the ternary operation, clean code, and your PR description is really good. I also notice you received feedback from a peer and improved your code, that’s the right mindset.

 Good job! 👍 

Evaluations coming later

Biffenx:
Thanks Jordane for your feedback. I know you are putting in a lot of time checking the codes and doing the feedbacks but need to say that thanks to the feedback from you I really enjoy to do the next challenges. 

Also I see how important it is for me to have access to my fellow peers and so proud to have great peers who takes their time to help and advice me who needs a bit more help….

Stay strong & 1000 tack!

Jack

PR #40, opened 2020-03-24
My solution

lsjoberg98:
Getting close to deadline


jgengo:
Unfortunately, you submitted your PR close to the deadline, so I guess you did not really have time to write a describing description of your code.

You include unistd.h, stdio.h, and stdlib.h but you are not using them. You should avoid including unused libs.

 Your code could be a little bit cleaner with few edits, I guess you were in the rush. Still, looks ok for me.  

Regarding the fact the description isn’t good enough, I close this PR.  

Good job though! 👍

PR #41, opened 2020-03-24
mdesta chall00

mdesta:
solution to chall00 using strcmp and strlen


mdesta:
chall00

jgengo:
Hello, you have pushed after the deadline. Sorry.

PR #43, opened 2020-03-26
Morse code

kimsappi:
Added usage printing and exit for special cases (no argument, too many arguments, empty string as argument, special characters)

I loop through the string, converting all alphas to lowercase and fetching the matching Morse code from the array charset, adding it to the output string. If the character is a space, a space is added instead. At the end the output string is printed.

### DIFF
--- a/chall01/ksappi.py
+++ b/chall01/ksappi.py
@@ -1,0 +1,18 @@
+#!/usr/bin/python3
+import sys
+
+charset = [".-", "-...", "-.-.", "-..", ".", "..-.", "--.", "....", "..",
+	".---", "-.-", ".-..", "--", "-.", "---", ".--.", "--.-", ".-.",
+	"...", "-", "..-", "...-", ".--", "-..-", "-.--", "--.."]
+
+def main():
+	output = ""
+	for char in sys.argv[1]:
+		if char.isalpha():
+			output += charset[ord(char.lower()) - ord('a')]
+		elif char == " ":
+			output += " "
+	print(output)
+
+if __name__ == "__main__":
+	main()
--- a/chall01/ksappi.py
+++ b/chall01/ksappi.py
@@ -7,13 +7,17 @@ charset = [".-", "-...", "-.-.", "-..", ".", "..-.", "--.", "....", "..",
 
 def main():
 	output = ""
-	if len(sys.argv) < 2:
+	if len(sys.argv) != 2 or len(sys.argv[1]) == 0:
+		print("usage: ./ksappi.py <a-zA-Z string>")
 		return
 	for char in sys.argv[1]:
 		if char.isalpha():
 			output += charset[ord(char.lower()) - ord('a')]
 		elif char == " ":
 			output += " "
+		else:
+			print("usage: ./ksappi.py <a-zA-Z string>")
+			return
 	print(output)
 
 if __name__ == "__main__":
--- a/chall01/ksappi.py
+++ b/chall01/ksappi.py
@@ -7,6 +7,8 @@ charset = [".-", "-...", "-.-.", "-..", ".", "..-.", "--.", "....", "..",
 
 def main():
 	output = ""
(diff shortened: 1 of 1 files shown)


salahadawi:
Throws an error with characters such as 'ä', 'ö', 'å'.
Isalpha() counts these as valid alphabetical characters, which is a problem since we don't have morse codes for them.

chall01/ksappi.py
```diff
@@ -0,0 +1,24 @@
...
+			output += charset[ord(char.lower()) - ord('a')]
+		elif char == " ":
+			output += " "
+		else:
+			print("usage: ./ksappi.py <a-zA-Z string>")
+			return
+	print(output)
+
+if __name__ == "__main__":
+	main()
```
d15ky, line 24: that's a good practice (protecting your script from running on the import), well done! 
however, don't forget about the newline at the end of the file, that's not a good practice but good manners :) And as you probably know, "manners maketh man".

chall01/ksappi.py
```diff
@@ -0,0 +1,24 @@
...
+import sys
+
+charset = [".-", "-...", "-.-.", "-..", ".", "..-.", "--.", "....", "..",
+	".---", "-.-", ".-..", "--", "-.", "---", ".--.", "--.-", ".-.",
+	"...", "-", "..-", "...-", ".--", "-..-", "-.--", "--.."]
+
+def main():
+	output = ""
+	if len(sys.argv) != 2 or len(sys.argv[1]) == 0:
+		print("usage: ./ksappi.py <a-zA-Z string>")
```
d15ky, line 11: it's better to use sys.argv[0] instead of hardcoding name of the program

chall01/ksappi.py
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/python3
+import sys
+
+charset = [".-", "-...", "-.-.", "-..", ".", "..-.", "--.", "....", "..",
+	".---", "-.-", ".-..", "--", "-.", "---", ".--.", "--.-", ".-.",
+	"...", "-", "..-", "...-", ".--", "-..-", "-.--", "--.."]
+
+def main():
+	output = ""
```
d15ky, line 9: small note:
just a good practice to initialize a variable you'll use in the loop right before the loop and not in the begging of function
especially if this variable will only be modified inside a loop, so it's kind of a "loop" variable, and if it's initialized somewhere else it's hard to track

chall01/ksappi.py
```diff
@@ -0,0 +1,24 @@
...
+		print("usage: ./ksappi.py <a-zA-Z string>")
+		return
+	for char in sys.argv[1]:
+		if char.isalpha():
+			output += charset[ord(char.lower()) - ord('a')]
+		elif char == " ":
+			output += " "
+		else:
+			print("usage: ./ksappi.py <a-zA-Z string>")
+			return
```
d15ky, line 20: With Python there is a so-called "pythonic" way to approach tasks. I won't explain it in details here, I'll just show how this loop can be done in "pythonic" way: 
`output = "".join(charset[ord(c) - ord('a')] if c != " " else " " for c in sys.argv[1].lower())`
and if you want to make sure that everything in the string is alpha or spaces, you can do it like that:
```
if not sys.argv[1].replace(" ", "").isalpha():
    print(usage)
    return
```
you can actually just add it to the "if" above and therefore all the logic of your program would be a oneliner
`print("".join(charset[ord(c) - ord('a')] if c != " " else " " for c in sys.argv[1].lower()))`

all of this is not to say that you've done something wrong but to show perks of Python and simplicity/laconicism it provides
the main reason why Python is used so widely in research, ML, mathematics, DA, etc. is not that it's a great language even tho it is but that it allows to code and test different hypothesis super fast. and if you see that your hypothesis works, you start coding it in C++ to make it fast.

d15ky (approved):
Overall, great job discovering Python and I especially admire your thoughtful usage of main function

PR #44, opened 2020-03-26
Morse code

salahadawi:
I check if there is precisely 2 arguments, if not the usage is printed.
The morse code letters are saved in a list.
Then, checking the input string one character at a time, i convert it to an index that can be used to find the corresponding letter in the morse code list.
Once all characters have been converted, the resulting string is printed.
If a character is not an alphabetical character, the script stops and usage is printed.

### DIFF
--- a/chall01/sadawi.py
+++ b/chall01/sadawi.py
@@ -40,12 +40,13 @@ result = ""
 input = sys.argv[1]
 if not input:
 	error()
-for c in input:
-	c = c.upper()
-	if not 65 <= ord(c) <= 90 and c != " ":
+for char in input:
+	char = char.upper()
+	asciiValue = ord(char)
+	if not 65 <= asciiValue <= 90 and char != " ":
 		error()
-	if c == " ":
+	if char == " ":
 		result += " "
 	else:
-		result += format[ord(c) - 65]
+		result += format[asciiValue - 65]
 print (result)
--- a/chall01/sadawi.py
+++ b/chall01/sadawi.py
@@ -1,0 +1,51 @@
+#! /usr/bin/python3
+
+import sys
+
+format = """.-
+-... 
+-.-. 
+-.. 
+. 
+..-. 
+--. 
+.... 
+.. 
+.--- 
+-.- 
+.-.. 
+-- 
+-. 
+--- 
+.--. 
+--.- 
+.-. 
+... 
+- 
+..- 
+...- 
+.-- 
+-..- 
+-.-- 
+--..""".split()
+
+def error():
+	print("usage: ./xlogin.py <a-zA-Z string>")
+	sys.exit()
+
+if len(sys.argv) != 2:
+	error()
+
+result = ""
+input = sys.argv[1]
+if not input:
+	error()
+for c in input:
+	c = c.upper()
+	if not 65 <= ord(c) <= 90 and c != " ":
+		error()
+	if c == " ":
+		result += " "
+	else:
+		result += format[ord(c) - 65]
+print (result)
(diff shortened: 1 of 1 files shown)


chall01/sadawi.py
```diff
@@ -0,0 +1,51 @@
...
+--.- 
+.-. 
+... 
+- 
+..- 
+...- 
+.-- 
+-..- 
+-.-- 
+--..""".split()
```
d15ky, line 30: As a programmer, I approve and support, it's the fastest and smart decision. 
As a reviewer/reader of your code, I'd prefer to see it in one line tho :) e.g. cmd/ctrl + D in sublime allows to quickly do that

chall01/sadawi.py
```diff
@@ -0,0 +1,51 @@
...
+
+def error():
+	print("usage: ./sadawi.py <a-zA-Z string>")
+	sys.exit()
```
d15ky, line 34: Nice! Separate function for separate logic is always good. However, you can use sys.argv[0] instead of hardcoding program name

chall01/sadawi.py
```diff
@@ -0,0 +1,51 @@
...
+
+if len(sys.argv) != 2:
+	error()
+
```
d15ky, line 38: A good practice is to protect your python scripts that are meant to be run from running on import. 
It's done by
`if __name__ == "__main__":`

chall01/sadawi.py
```diff
@@ -0,0 +1,51 @@
...
+for char in sys.argv[1]:
+	char = char.upper()
+	asciiValue = ord(char)
+	if not 65 <= asciiValue <= 90 and char != " ":
+		error()
+	if char == " ":
+		result += " "
+	else:
+		result += format[asciiValue - 65]
+print (result)
```
d15ky, line 51: Newline at the end of the file is always nice. "Manners maketh man" as they say

chall01/sadawi.py
```diff
@@ -0,0 +1,51 @@
...
+--..""".split()
+
+def error():
+	print("usage: ./sadawi.py <a-zA-Z string>")
+	sys.exit()
+
+if len(sys.argv) != 2:
+	error()
+
+result = ""
```
d15ky, line 39: it's better to initialize your "loop" variable (that only used as a container for the loop execution result) right before the loop

chall01/sadawi.py
d15ky: Also "format" is the name of a built-in function, pay attention to that

chall01/sadawi.py
```diff
@@ -0,0 +1,51 @@
...
+	error()
+for char in sys.argv[1]:
+	char = char.upper()
+	asciiValue = ord(char)
+	if not 65 <= asciiValue <= 90 and char != " ":
+		error()
+	if char == " ":
+		result += " "
+	else:
+		result += format[asciiValue - 65]
```
d15ky, line 50: There is always a so-called "pythonic" way to write code when you do it in Python. I won't describe it in details but here is an example of how it can be done here: 
`result = "".join(format[ord(c) - 65] if c != " " else " " for c in sys.argv[1].upper())`
and if you want to check that every character in the string is alpha, you can do that before (where you check for errors):
```
if not sys.argv[1] or not sys.argv[1].replace(" ", "").isalpha():
    error()
```

d15ky: It's not like you've done something wrong, a lot of experienced programmers don't know where they can do something more "pythonic". But that's one of the main advantages of Python - it's quick to write Python code, and it's laconic. That's why it used so widely in the research field to test a hypothesis, and then, if the hypothesis works, they rewrite it in C++.

d15ky (approved):
Good job exploring Python and I like separate function for error.

PR #45, opened 2020-03-26
Morse Code

vossr:
First check errors, then loop input string with 27 if cases and print matching morse sequence.

### DIFF
--- a/chall01/rpehkone.py
+++ b/chall01/rpehkone.py
@@ -1,0 +1,68 @@
+#!/usr/bin/python3
+
+import sys
+
+if len(sys.argv) != 2 or not sys.argv[1]:
+	print("usage: " + sys.argv[0] + " <a-zA-Z string>")
+	exit(0)
+for i in sys.argv[1]:
+	if not i.isalnum() and i != ' ':
+		print("usage: " + sys.argv[0] + " <a-zA-Z string>")
+		exit(0)
+sys.argv[1] = sys.argv[1].lower()
+for i in sys.argv[1]:
+	if i == "a":
+		print(".-", end =""),
+	elif i == "b":
+		print("-...", end =""),
+	elif i == "c":
+		print("-.-.", end =""),
+	elif i == "d":
+		print("-..", end =""),
+	elif i == "e":
+		print(".", end =""),
+	elif i == "f":
+		print("..-.", end =""),
+	elif i == "g":
+		print("--.", end =""),
+	elif i == "h":
+		print("....", end =""),
+	elif i == "i":
+		print("..", end =""),
+	elif i == "j":
+		print(".---", end =""),
+	elif i == "k":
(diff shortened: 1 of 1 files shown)


chall01/rpehkone.py
```diff
@@ -0,0 +1,68 @@
+#!/usr/bin/python3
+
+import sys
+
+if len(sys.argv) != 2 or not sys.argv[1]:
+	print("usage: " + sys.argv[0] + " <a-zA-Z string>")
+	exit(0)
```
d15ky, line 7: nice! good check and well done using sys.argv[0]

chall01/rpehkone.py
```diff
@@ -0,0 +1,68 @@
...
+for i in sys.argv[1]:
+	if not (i.isalpha() and ord(i) < 128) and i != ' ':
+		print("usage: " + sys.argv[0] + " <a-zA-Z string>")
+		exit(0)
```
d15ky, line 11: well done using sys.argv[0] for the program name and checking string beforehand for errors
however, overall this check can be done in a more "pythonic" way: 
`if not sys.argv[1].replace(" ", "").isalpha():`

chall01/rpehkone.py
```diff
@@ -0,0 +1,68 @@
...
+sys.argv[1] = sys.argv[1].lower()
```
d15ky, line 12: that's a smart move too

chall01/rpehkone.py
```diff
@@ -0,0 +1,68 @@
...
+		print(".--", end =""),
+	elif i == "x":
+		print("-..-", end =""),
+	elif i == "y":
+		print("-.--", end =""),
+	elif i == "z":
+		print("--..", end =""),
+	else:
+		print(" ", end =""),
+print()
```
d15ky, line 68: Well, it's not the most pythonic way to accomplish this task
First, I'd use data given (all the codes) and just copy-paste it inside as an array:
`morse_codes = '.- -...  -.-.  -..  .  ..-.  --.  ....  ..  .---  -.-  .-..  --  -.  ---  .--.  --.-  .-.  ...  -  ..-  ...-  .--  -..-  -.--  --..'.split(' ')`
Then, let's make it a dict - this way it's prettier
`morse_codes = {chr(65 + i) : morse_codes[i] for i in range(len(morse_codes))}` 65 here is code of 'A'
And let's add to our dict a space sign too: 
`morse_codes[" "] = " "`
And now we can easily print out in morse code any string given: 
`print("".join(morse_codes[x] for x in sys.argv[1].upper()))`

Python code should be pythonic (laconic). That's the main advantage of Python, the possibility to write scripts fast and with less code.

d15ky (approved):
Overall, good! I liked your code. It could use some pythonicism (and 27 ifs is a little too much) but other than that, great job discovering Python!

PR #46, opened 2020-03-26
Morse code

osalmine:
First I check the argument amount, if it's not 2 the program exits. After this I take the first argument and put it to uppercase and check if it contains only alphabets and spaces. Finally I browse through the morse code dictionary and match each letter to the code and print it out.

### DIFF
--- a/chall01/osalmine.py
+++ b/chall01/osalmine.py
@@ -1,0 +1,33 @@
+#!/usr/bin/python3
+
+import sys
+
+morse = { 'A':'.-', 'B':'-...',
+                    'C':'-.-.', 'D':'-..', 'E':'.',
+                    'F':'..-.', 'G':'--.', 'H':'....',
+                    'I':'..', 'J':'.---', 'K':'-.-',
+                    'L':'.-..', 'M':'--', 'N':'-.',
+                    'O':'---', 'P':'.--.', 'Q':'--.-',
+                    'R':'.-.', 'S':'...', 'T':'-',
+                    'U':'..-', 'V':'...-', 'W':'.--',
+                    'X':'-..-', 'Y':'-.--', 'Z':'--..'}
+
+if len(sys.argv) != 2:
+	print("usage: ./osalmine.py <a-zA-Z string>")
+	quit()
+
+string = sys.argv[1].upper();
+if len(string) == 0:
+	print("usage: ./osalmine.py <a-zA-Z string>")
+	quit()
+
+if not all(x.isalpha() or x.isspace() for x in string):
+	print("usage: ./osalmine.py <a-zA-Z string>")
+	quit()
+
+for letter in string:
+	if (letter == " "):
+		print(" ", end = '')
+	else:
+		print(morse[letter], end = '');
+print()


chall01/osalmine.py
```diff
@@ -0,0 +1,33 @@
...
+
+morse = { 'A':'.-', 'B':'-...',
+                    'C':'-.-.', 'D':'-..', 'E':'.',
+                    'F':'..-.', 'G':'--.', 'H':'....',
+                    'I':'..', 'J':'.---', 'K':'-.-',
+                    'L':'.-..', 'M':'--', 'N':'-.',
+                    'O':'---', 'P':'.--.', 'Q':'--.-',
+                    'R':'.-.', 'S':'...', 'T':'-',
+                    'U':'..-', 'V':'...-', 'W':'.--',
+                    'X':'-..-', 'Y':'-.--', 'Z':'--..'}
```
d15ky, line 13: That's I approve! Smart! 
One little suggestion (in case you hardcoded it), you could just use codes given and make dict out of them using python like that: 
```
morse = '.- -... -.-. -.. . ..-. --. .... .. .--- -.- .-.. -- -. --- .--. --.- .-. ... - ..- ...- .-- -..- -.-- --..'.split(' ')
morse = {chr(65 + i) : morse[i] for i in range(len(morse))} # 65 is code of 'A'
```
But that's not all! You can also add a space as a key to your dictionary instead of checking it in the printing loop:
`morse[" "] = " "`
and that's it :)

chall01/osalmine.py
```diff
@@ -0,0 +1,33 @@
...
+
+if len(sys.argv) != 2:
+	print("usage: ./osalmine.py <a-zA-Z string>")
+	quit()
+
+string = sys.argv[1].upper();
```
d15ky, line 19: changing string to upper - smart! 
also, I see C-legacy here ';' ;) 
but yeah, I can relate, it's not a deal

chall01/osalmine.py
```diff
@@ -0,0 +1,33 @@
...
+if len(string) == 0:
+	print("usage: ./osalmine.py <a-zA-Z string>")
+	quit()
+
+if not all(x.isalpha() or x.isspace() for x in string):
+	print("usage: ./osalmine.py <a-zA-Z string>")
+	quit()
```
d15ky, line 26: All the checks for errors could be in one if or in one function. This way it's a little hard to read and also it breaks the main rule of programming: DRY (Don't Repeat Yourself)

Also, exactly in this check, you use the function `isspace()` which recognizes all kinds of spaces but later on you only check for one space sign. If someone will use '\t', your script will crash.

chall01/osalmine.py
```diff
@@ -0,0 +1,33 @@
...
+
+for letter in string:
+	if (letter == " "):
+		print(" ", end = '')
+	else:
+		print(morse[letter], end = '');
```
d15ky, line 32: There is a so-called pythonic way to write code in Python. Which in practice means using all the tools provided to write laconic and readable code. You can read more about pythonic yourself, I'll only provide an example in this particular case:
`print(*(morse[letter] for letter in string), sep="")`
works if you added space to keys of morse. if not, would look a bit more ugly:
`print(*(morse[letter] if letter != " " else " " for letter in string), sep="")`

d15ky (approved):
Very easy to read and good-looking solution overall. There are some minor suggestions from me but nothing crucial. Well done!

PR #47, opened 2020-03-26
Add solution

hiiri:
First check if argument is valid (letters or spaces).
Morse codes are sorted by alphabetical order already, so parse them into a dictionary with corresponding letters.
Then iterate over argument, printing morse code characters corresponding to each letter.

### DIFF
--- a/chall01/alcohen.py
+++ b/chall01/alcohen.py
@@ -28,14 +28,22 @@ c = """.-
 -..- 
 -.-- 
 --.."""
-string_alpha_or_spaces = all(x.isalpha() or x.isspace() for x in sys.argv[1])
-if (len(sys.argv) != 2 or not string_alpha_or_spaces or not sys.argv[1]):
+
+print(sys.argv[1])
+def error():
     print("usage: ./alcohen.py <a-zA-Z string>")
     exit()
+
+if (len(sys.argv) != 2):
+    error()
+
+string_alpha_or_spaces = all(x.isalpha() or x.isspace() for x in sys.argv[1])
+if (not string_alpha_or_spaces or not sys.argv[1]):
+    error()
+
 codes = c.split()
 alph = list(string.ascii_lowercase)
 d = dict(zip(alph, codes))
-
 
 for x in str(sys.argv[1]).lower():
     if (x == " "):
--- a/chall01/alcohen.py
+++ b/chall01/alcohen.py
@@ -1,4 +1,4 @@
-mport string
+import string
 import sys
 
 c = """.-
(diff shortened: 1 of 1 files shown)


chall01/alcohen.py
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env python3
+import string
+import sys
+
+def error():
+    print("usage: ./alcohen.py <a-zA-Z string>")
+    exit()
```
jgengo, line 7: I would suggest using sys.argv[0] for avoiding hardcoding alcohen into your print.

chall01/alcohen.py
```diff
@@ -0,0 +1,30 @@
...
+
+# Morse codes in order from A to Z
+morse_codes = ['.-', '-...', '-.-.', '-..', '.', '..-.', '--.', '....', '..', '.---', '-.-', '.-..', '--', '-.', '---', '.--.', '--.-', '.-.', '...', '-', '..-', '...-', '.--', '-..-', '-.--', '--..']
+
+if (len(sys.argv) != 2): # Only accept one argument
+    error()
+
+# alphabet contains letters [a-z] and a space
+alphabet = list(string.ascii_lowercase + ' ')
```
jgengo, line 16: really good!

chall01/alcohen.py
```diff
@@ -0,0 +1,30 @@
...
+string_to_convert = sys.argv[1].lower()
+
+# Check that string to convert isn't empty and only contains characters that are in alphabet.
+string_is_a_to_z_or_spaces = all(letter in alphabet for letter in string_to_convert) 
+if (not string_is_a_to_z_or_spaces or not string_to_convert):
+    error()
+
+# Combine letters and morse codes to a dictionary. A letter can be looked up to receive the corresponding morse code.
+alphabet_morse_dict = dict(zip(alphabet, morse_codes))
```
jgengo, line 25: really good usage of dict and zip

chall01/alcohen.py
```diff
@@ -0,0 +1,30 @@
...
+alphabet_morse_dict[' '] = ' '
+
+for letter in string_to_convert:
+        print(alphabet_morse_dict[letter], end='')
```
jgengo, line 29: I would suggest appending into a string and then use print only once.

chall01/alcohen.py
```diff
@@ -0,0 +1,30 @@
...
+print()
```
jgengo, line 30: A good practice is to protect your python scripts that are meant to be run from running on import.
It's done by
```python
if __name__ == "__main__":
```

PR #48, opened 2020-03-26
Morse Code (jnovotny)

jiricodes:
The script checks if amount of arguments is correct and whether the given string contains only alphabetical symbols or space. If not, the exit message is displayed. Otherwise the string is converted into Morse code using static dictionary and printed out.


chall01/jnovotny.py
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/python3
+import sys
+
+def usage_exit():
+	print("usage: ./jnovotny.py <a-zA-Z string>")
```
jgengo, line 5: I would suggest avoiding to hardcore jnovotny and favorize `sys.argv[0]`

chall01/jnovotny.py
```diff
@@ -0,0 +1,52 @@
...
+		"q":"--.-",
+		"r":".-.",
+		"s":"...",
+		"t":"-",
+		"u":"..-",
+		"v":"...-",
+		"w":".--",
+		"x":"-..-",
+		"y":"-.--",
+		"z":"--.."}
```
jgengo, line 35: in python, I would recommend putting this dict outside of the function, so if you use many times morse you don't declare this again.

chall01/jnovotny.py
```diff
@@ -0,0 +1,52 @@
...
+	str_out = ""
+	for letter in string:
+		if letter == " ":
+			str_out += " "
+		elif letter in morse_dict.keys():
+			str_out += f"{morse_dict[letter]}"
```
jgengo, line 41: why do you format it?

str_out += morse_dict[letter] should works fine.

chall01/jnovotny.py
```diff
@@ -0,0 +1,52 @@
...
+		"v":"...-",
+		"w":".--",
+		"x":"-..-",
+		"y":"-.--",
+		"z":"--.."}
+	str_out = ""
+	for letter in string:
+		if letter == " ":
+			str_out += " "
+		elif letter in morse_dict.keys():
```
jgengo, line 40: I would suggest adding space into your dict and reformat this with
```
str_out += morse_dict[letter]
```
and put the `else` out of here as an error case handling before going inside this loop.

chall01/jnovotny.py
```diff
@@ -0,0 +1,52 @@
...
+			str_out += f"{morse_dict[letter]}"
+		else:
+			usage_exit()
+	return str_out
+
+
+
+if __name__ == "__main__":
```
jgengo, line 48: really good!

PR #49, opened 2020-03-26
Add ohakola chall01

hakolao:
Done!

### DIFF
--- a/chall01/ohakola.py
+++ b/chall01/ohakola.py
@@ -1,0 +1,53 @@
+#!/usr/bin/env python3
+import sys
+
+usage = "usage: ./ohakola.py <a-zA-Z string>"
+morse_dict = {
+	"a": ".-",
+	"b": "-...",
+	"c": "-.-.",
+	"d": "-..",
+	"e": ".",
+	"f": "..-.",
+	"g": "--.",
+	"h": "....",
+	"i": "..",
+	"j": ".---",
+	"k": "-.-",
+	"l": ".-..",
+	"m": "--",
+	"n": "-.",
+	"o": "---",
+	"p": ".--.",
+	"q": "--.-",
+	"r": ".-.",
+	"s": "...",
+	"t": "-",
+	"u": "..-",
+	"v": "...-",
+	"w": ".--",
+	"x": "-..-",
+	"y": "-.--",
+	"z": "--.."
+}
+
+def print_args():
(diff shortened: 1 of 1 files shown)


chall01/ohakola.py
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env python3
+import sys
+
+usage = "usage: ./ohakola.py <a-zA-Z string>"
```
jgengo, line 4: nice! but don't hardcode your app name when you shouldn't use `sys.argv[0]`

chall01/ohakola.py
```diff
@@ -0,0 +1,53 @@
...
+			char = str[i]
+			if (i < str_len - 1):
+				print(morse_dict[char], end = '')
+			else:
+				print(morse_dict[char], end = '\n')
+			i += 1
+	else:
+		print(usage)
+
+print_args()
```
jgengo, line 53: A good practice is to protect your python scripts that are meant to be run from running on import.
It's done by
```python
if __name__ == "__main__":
  print_args()
```

chall01/ohakola.py
```diff
@@ -0,0 +1,53 @@
...
+	"w": ".--",
+	"x": "-..-",
+	"y": "-.--",
+	"z": "--.."
+}
+
+def print_args():
+	args_len = len(sys.argv)
+	str = sys.argv[1].lower()
+	if (args_len == 2):
```
jgengo, line 37: I would recommend checking your args length outside of this function

chall01/ohakola.py
```diff
@@ -0,0 +1,53 @@
...
+		str_len = len(str)
+		if (str_len == 0):
+			print(usage)
+			return
```
jgengo, line 41:
```python
if not str:
```

chall01/ohakola.py
```diff
@@ -0,0 +1,53 @@
...
+		i = 0
+		while (i < str_len):
+			char = str[i]
+			if (i < str_len - 1):
+				print(morse_dict[char], end = '')
```
jgengo, line 46: You should append into a string and then print at the end to avoid using print for every character.

jgengo (approved):

chall01/ohakola.py
hakolao: Thanks so much for comments! Good points all :)

PR #50, opened 2020-03-26
submit chall01 solution

tdawson-42:
First I check the amount of arguments is correct and that the input is a string that is at least 1 character long. Then I create a dictionary with matching letter : morse code pairs. Case doesn't matter with morse code so we make the string all lowercase. I then check that the string contains only characters that are keys in the dictionary (a-z and space).  We can then iterate through the string one character at a time and print the morse code equivalent for each character.

### DIFF
--- a/chall01/tdawson.py
+++ b/chall01/tdawson.py
@@ -1,17 +1,9 @@
 #!/usr/bin/python3
 import sys
 
-if len(sys.argv) != 2:
+if len(sys.argv) != 2 or not isinstance(sys.argv[1], str) or len(sys.argv[1]) == 0:
 	print("usage: ./xlogin.py <a-zA-Z string>")
 	sys.exit()
-
-string = sys.argv[1]
-
-if not all(c.isalpha() or c.isspace() for c in string) or len(string) == 0:
-	print("usage: ./xlogin.py <a-zA-Z string>")
-	sys.exit()
-
-string = string.lower()
 
 morse_alphabet = {
 	"a" : ".-",
@@ -42,6 +34,12 @@ morse_alphabet = {
 	"z" : "--..",
 	" " : " "
 }
+
+string = sys.argv[1].lower()
+
+if not all(c in morse_alphabet for c in string):
+	print("usage: ./xlogin.py <a-zA-Z string>")
+	sys.exit()
 
 for c in string:
 	print(morse_alphabet[c], end="")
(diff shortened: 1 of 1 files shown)


chall01/tdawson.py
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/python3
+import sys
+
+if len(sys.argv) != 2 or len(sys.argv[1]) == 0:
```
jgengo, line 4:
```python
len(sys.argv[1]) == 0
```
can be written
```python
or not sys.argv[1]
```

chall01/tdawson.py
```diff
@@ -0,0 +1,20 @@
...
+    sys.exit(f'usage: {sys.argv[0]} <a-zA-Z string>')
```
jgengo, line 5: really good! really!

chall01/tdawson.py
```diff
@@ -0,0 +1,20 @@
...
+
+morse_alphabet = {'a': '.-', 'b': '-...', 'c': '-.-.', 'd': '-..', 'e': '.',
```
jgengo, line 7: once again really good!

chall01/tdawson.py
```diff
@@ -0,0 +1,20 @@
...
+                  'f': '..-.', 'g': '--.', 'h': '....', 'i': '..', 'j': '.---',
+                  'k': '-.-', 'l': '.-..', 'm': '--', 'n': '-.', 'o': '---', 'p': '.--.',
+                  'q': '--.-', 'r': '.-.', 's': '...', 't': '-', 'u': '..-', 'v': '...-',
+                  'w': '.--', 'x': '-..-', 'y': '-.--', 'z': '--..', ' ': ' '}
+
+string = sys.argv[1].lower()
+
+if not all(c in morse_alphabet for c in string):
+    sys.exit(f'usage: {sys.argv[0]} <a-zA-Z string>')
```
jgengo, line 16: good!

chall01/tdawson.py
```diff
@@ -0,0 +1,20 @@
...
+
+for c in string:
+    print(morse_alphabet[c], end='')
```
jgengo, line 19: I would suggest avoiding printing char by char but better to append into a string and then print the string at the end.

tdawson-42: Hi, thank you for your feedback, it is very much appreciated! 👍 

I agree with your suggestion here, I learned about the `String.join()` method so I would now do it this way:
```Python
print(''.join(morse_alphabet[c] for c in string))
```

PR #51, opened 2020-03-26
Morse code

vkuikka:
Finds the index of each given character in the alphabet and prints the corresponding code.
I used the index because vim could easily print all numbers at once and I did not have to write them in each if statement.

### DIFF
--- a/chall01/vkuikka.py
+++ b/chall01/vkuikka.py
@@ -1,63 +1,71 @@
 #!/usr/bin/python3
 import sys
 
-alphabet = "abcdefghijklmnopqrstuvwxyz"
+def my_errorMessage():
+	print ("usage:", sys.argv[0], "<a-zA-Z string>")
+	quit()
+
+alphabet = "abcdefghijklmnopqrstuvwxyz "
+morse = ""
 
 if len(sys.argv) != 2:
-	print ("usage: ./vkuikka.py <a-zA-Z string")
-	quit()
+	my_errorMessage()
 for arg_ch in sys.argv[1]:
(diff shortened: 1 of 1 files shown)


chall01/vkuikka.py
```diff
@@ -0,0 +1,71 @@
+#!/usr/bin/python3
+import sys
+
+def my_errorMessage():
+	print ("usage:", sys.argv[0], "<a-zA-Z string>")
+	quit()
```
jgengo, line 6: I would suggest changing your print by :

```python
```

Plus the good practice about printing usage is : 
- if it's due to an input/user error --> stderr + return something else than 0
- if it's requested by the user (e.g `./app.py --help`) --> stdout + return 0

chall01/vkuikka.py
```diff
@@ -0,0 +1,71 @@
...
+	elif alpha_i == 23:
+		morse += "-..-"
+	elif alpha_i == 24:
+		morse += "-.--"
+	elif alpha_i == 25:
+		morse += "--.."
+	elif alpha_i == 26:
+		morse += " "
+	else:
+		my_errorMessage()
```
jgengo, line 70: If forest ... 

I guess it works but it's not really efficient code

You could have use a python dict for example and just do `morse += mydict[alpha_i] `

see here: https://www.tutorialspoint.com/python3/python_dictionary.htm

jgengo (approved):
You should avoid If-forest. Even if it works, it's important for you to push yourself to discover some better ways to do that in Python.

It's done by
```python
if __name__ == "__main__":
```

PR #52, opened 2020-03-26
Add krusthol's solution for String to Morse

krusthol:
Approach:
Due to very little Python writing experience, a simple & naive, straight-forward solution was written. I simply thought of the capabilities of Python to be atleast what I could write with in C, so I looked for a C style solution, which results in a sub-optimal solution when compared to what you could truly do with Python. Some Python-like elements were used, mostly quickly referenced from W3School and GeeksForGeeks.

Code:
Simple Python3 executable script that encodes string to morse, using 'import sys' and 'import string' python3 libs for checking that argument list len equals 2 (one argument supplied) and that one supplied argument is within bounds of a-z, A-Z, ' ' (space). For printing morse, uses a if, else if -tree that iterates and checks against the lowercased characters of the supplied argument. Succesful check stdout writes characters to respective morse, but prints spaces as spaces. Script finishes by printing a newline and exits with sys success code 0.

### DIFF
--- a/chall01/krusthol.py
+++ b/chall01/krusthol.py
@@ -12,7 +12,7 @@ def bad_character(str):
 if len(sys.argv) != 2:
     sys.exit("usage: ./krusthol.py <a-zA-Z string>")
 arg = sys.argv[1].lower()
-if bad_character(arg):
+if len(arg) == 0 or bad_character(arg):
     sys.exit("usage: ./krusthol.py <a-zA-Z string>")
 for c in arg:
     if c == 'a':
--- a/chall01/krusthol.py
+++ b/chall01/krusthol.py
@@ -1,0 +1,73 @@
+#!/usr/bin/python3
+import sys
+import string
+
+def bad_character(str):
+    good = set(string.ascii_lowercase + ' ')
+    for c in str:
+        if c not in good:
+            return True
+    return False
+
+if len(sys.argv) != 2:
+    sys.exit("usage: ./krusthol.py <a-zA-Z string>")
+arg = sys.argv[1].lower()
+if bad_character(arg):
+    sys.exit("usage: ./krusthol.py <a-zA-Z string>")
+for c in arg:
+    if c == 'a':
+        sys.stdout.write('.-')
+    elif c == 'b':
+        sys.stdout.write('-...')
+    elif c == 'c':
+        sys.stdout.write('-.-.')
+    elif c == 'd':
+        sys.stdout.write('-..')
+    elif c == 'e':
+        sys.stdout.write('.')
+    elif c == 'f':
+        sys.stdout.write('..-.')
+    elif c == 'g':
+        sys.stdout.write('--.')
+    elif c == 'h':
+        sys.stdout.write('....')
+    elif c == 'i':
+        sys.stdout.write('..')
+    elif c == 'j':
+        sys.stdout.write('.---')
+    elif c == 'k':
+        sys.stdout.write('-.-')
+    elif c == 'l':
+        sys.stdout.write('.-..')
+    elif c == 'm':
+        sys.stdout.write('--')
+    elif c == 'n':
+        sys.stdout.write('-.')
+    elif c == 'o':
+        sys.stdout.write('---')
+    elif c == 'p':
+        sys.stdout.write('.--.')
+    elif c == 'q':
+        sys.stdout.write('--.-')
+    elif c == 'r':
+        sys.stdout.write('.-.')
+    elif c == 's':
+        sys.stdout.write('...')
+    elif c == 't':
+        sys.stdout.write('-')
+    elif c == 'u':
+        sys.stdout.write('..-')
+    elif c == 'v':
+        sys.stdout.write('...-')
+    elif c == 'w':
+        sys.stdout.write('.--')
+    elif c == 'x':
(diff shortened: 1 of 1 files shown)


chall01/krusthol.py
```diff
@@ -0,0 +1,73 @@
+#!/usr/bin/python3
+import sys
+import string
+
+def bad_character(str):
+    good = set(string.ascii_lowercase + ' ')
+    for c in str:
+        if c not in good:
+            return True
+    return False
```
d15ky, line 10: I always admire creating separate functions for separate functions (sorry for tautology but it seems to me that people tend to neglect the fact that "function" is not randomly picked word for programming term, it literally mean ONE FUNCTION / ONE PURPOSE / etc. and that means that if you have 1 part of a code that serves one function it should be separated to one function :D) in code logic. So great job here! 
Also nice usage of "string" library and "set" file type.

chall01/krusthol.py
```diff
@@ -0,0 +1,73 @@
...
+
+if len(sys.argv) != 2:
```
d15ky, line 12: Would be nice to put all the error checks to one "if" statement since otherwise you kind of break the rule of DRY (don't repeat yourself) - one of the most important rules of programming

chall01/krusthol.py
```diff
@@ -0,0 +1,73 @@
...
+    sys.exit("usage: ./krusthol.py <a-zA-Z string>")
```
d15ky, line 13: Well done using the possibility to pass the error message to the "exit" function! 
One advice - it's better not to hardcode the name of your program in the error message but use sys.argv[0]. You can do so in C as well.

chall01/krusthol.py
```diff
@@ -0,0 +1,73 @@
...
+arg = sys.argv[1].lower()
```
d15ky, line 14: changing the whole string to lowercase - smart!

chall01/krusthol.py
```diff
@@ -0,0 +1,73 @@
...
+    elif c == 'w':
+        sys.stdout.write('.--')
+    elif c == 'x':
+        sys.stdout.write('-..-')
+    elif c == 'y':
+        sys.stdout.write('-.--')
+    elif c == 'z':
+        sys.stdout.write('--..')
+    elif c == ' ':
+        sys.stdout.write(' ')
```
d15ky, line 71: Here goes my main advice and a short introduction to "pythonic" way to write code in Python. 
First of all, instead of using numerous if-else statements, it's better to use dictionaries. Here is an example:
`my_awesome_dict = {'a': '.-', 'b': '-...'}`
Dictionary allows creating key-value pairs which is invaluably useful in so many cases. 
And here how you can quickly create one you need in pythonic way:
```
morse_input = ".- -..." # if you have multiple line string, you can create one in python using """ - triple quotes
morse_array = morse_input.split(" ") # again, if every value on new string, you can use "\n" instead of " ". And actually it's not array - in python it called list but it's kind of similar to C array
morse_dictionary = dict(zip(string.ascii_lowercase, morse_array) # this will create a dictionary in which every letter related to corresponding morse code
# or
morse_dictionary = {string.ascii_lowercase[i]: morse_array[i] for i in range(len(string.ascii_lowercase))}
# then add a space
morse_dictionary[" "] = " "
```
now we have a full dictionary
after that, all the if else statements can be truncated to one line:
`sys.stdout.write(morse_dictionary[c])`
even more than that! In Python, you can do the whole loop in one line by creating a string to print:
`print("".join(morse_dictionary[c] for c in arg))` 

And that is how laconic-pythonic Python is!

chall01/krusthol.py
```diff
@@ -0,0 +1,73 @@
...
+    return False
+
+if len(sys.argv) != 2:
+    sys.exit("usage: ./krusthol.py <a-zA-Z string>")
+arg = sys.argv[1].lower()
+if len(arg) == 0 or bad_character(arg):
+    sys.exit("usage: ./krusthol.py <a-zA-Z string>")
+for c in arg:
+    if c == 'a':
+        sys.stdout.write('.-')
```
d15ky, line 19: In python3 you can use print statement without a newline:
`print('.-', end='')`

chall01/krusthol.py
```diff
@@ -0,0 +1,73 @@
...
+    elif c == 'x':
+        sys.stdout.write('-..-')
+    elif c == 'y':
+        sys.stdout.write('-.--')
+    elif c == 'z':
+        sys.stdout.write('--..')
+    elif c == ' ':
+        sys.stdout.write(' ')
+print("")
+sys.exit(0)
```
d15ky, line 73: Don't forget about the newline! :) "Manners maketh man" as they say

d15ky (approved):
Hey, I want to overall note your amazing job in discovering Python and outstanding PR description. Suggestions below most likely will be some higher-level stuff and something you didn't encounter in C yet but I think that it still will be useful for you to think about or get to know. E.g. you may get interested in dictionaries and explore how to implement them in C-lang :)

krusthol:
Thanks a ton for the detailed feedback on the PR! In just a few short days, reading this feedback and the feedback for other pull requests in the challenge, I feel I got an expanded image of what the capabilities of Python are, and how one could more effortlessly express things with Python code. Remote-challs format thus seems effective and worthwhile. Thank you!

PR #54, opened 2020-03-26
working solution

avocadohooman:
My first ever written code in python :)

Created a dictionary (CODE) which includes all the morse code translations 
If received number of arguments is not 2, print usage.
If one of the letters in the received string is not part of the dictionary (a-z 0-9), print usage.
Otherwise print for each letter in the received string the morse value in the CODE dictionary.

### DIFF
--- a/chall01/gmolin.py
+++ b/chall01/gmolin.py
@@ -1,0 +1,40 @@
+#!/usr/bin/python
+import sys
+
+CODE = {'A': '.-',     'B': '-...',   'C': '-.-.', 
+        'D': '-..',    'E': '.',      'F': '..-.',
+        'G': '--.',    'H': '....',   'I': '..',
+        'J': '.---',   'K': '-.-',    'L': '.-..',
+        'M': '--',     'N': '-.',     'O': '---',
+        'P': '.--.',   'Q': '--.-',   'R': '.-.',
+     	'S': '...',    'T': '-',      'U': '..-',
+        'V': '...-',   'W': '.--',    'X': '-..-',
+        'Y': '-.--',   'Z': '--..',
+        
+        '0': '-----',  '1': '.----',  '2': '..---',
+        '3': '...--',  '4': '....-',  '5': '.....',
+        '6': '-....',  '7': '--...',  '8': '---..',
+        '9': '----.',  ' ': ' ',
+        }
+
+def main():
+    nb = len(sys.argv)
+
+    if nb == 2:
+        msg = sys.argv[1]
+        if msg == "":
+            print "usage: ./gmolin.py <a-zA-Z string>"
+            sys.exit()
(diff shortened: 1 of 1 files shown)


avocadohooman:
Created a dictionary (CODE)
If argv is not 2, print usage.
If one of the received letter in the string is not part of the dictionary (a-z 0-9), print usage.
Otherwise print for each letter in the received string the stored dictionary.

chall01/gmolin.py
```diff
@@ -0,0 +1,40 @@
...
+            else:
+                print "usage: ./gmolin.py <a-zA-Z string>"
+                sys.exit()
+        for char in msg:
+            print CODE[char.upper()],
+    else:
+        print "usage: ./gmolin.py <a-zA-Z string>"
+
+if __name__ == "__main__":
+	main()
```
jgengo, line 40: Yes! Good, you protected it!

Don't forget to add a new empty line at the end of your code.

chall01/gmolin.py
```diff
@@ -0,0 +1,40 @@
...
+    if nb == 2:
+        msg = sys.argv[1]
+        if msg == "":
+            print "usage: ./gmolin.py <a-zA-Z string>"
+            sys.exit()
+        for char in msg:
+            if char.upper() in CODE:
+                continue
+            else:
+                print "usage: ./gmolin.py <a-zA-Z string>"
```
jgengo, line 32: Don't hard code your name favorize `sys.argv[0]` if your code becomes bigger and one day you decide to call it another way you would have to edit your whole code.

Plus, I would also recommend creating a function print_usage() for the DRY (don't repeat yourself) concept

jgengo, line 32:
```python
if not msg.replace(" ", "").isalpha():
  print_usage()
```

chall01/gmolin.py
```diff
@@ -0,0 +1,40 @@
...
+                sys.exit()
+        for char in msg:
+            print CODE[char.upper()],
```
jgengo, line 35: Like in C, you should avoid print char by char but print a whole string in once.

chall01/gmolin.py
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/python
+import sys
+
+CODE = {'A': '.-',     'B': '-...',   'C': '-.-.',
```
jgengo, line 4: I suggest googling `PEP8` it can help with the variable convention name, `code` should be lowercase.

chall01/gmolin.py
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/python
```
jgengo, line 1: #!/usr/bin/python3

jgengo (approved):

PR #55, opened 2020-03-26
My Solution for 01

tuommii:
Checking with regex, that no other characters than a-Z or space are given. Then encrypting one character at time.

### DIFF
--- a/chall01/mtuomine.py
+++ b/chall01/mtuomine.py
@@ -1,0 +1,51 @@
+#!/usr/bin/env python3
+
+import sys
+import re
+
+MORSES = {
+'A':'.-',
+'B':'-...',
+'C':'-.-.',
+'D':'-..',
+'E':'.',
+'F':'..-.',
+'G':'--.',
+'H':'....',
+'I':'..',
+'J':'.---',
+'K':'-.-',
+'L':'.-..',
+'M':'--',
+'N':'-.',
+'O':'---',
+'P':'.--.',
+'Q':'--.-',
+'R':'.-.',
+'S':'...',
+'T':'-',
+'U':'..-',
+'V':'...-',
+'W':'.--',
+'X':'-..-',
+'Y':'-.--',
+'Z':'--..'
+}
+
+def encrypt(msg):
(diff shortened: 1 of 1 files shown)


chall01/mtuomine.py
```diff
@@ -0,0 +1,51 @@
...
+	msg = msg.upper()
+	morse = ''
+	for c in msg:
+		if c == ' ':
+			morse += ' '
+		else:
+			morse += MORSES[c]
+	return morse
+
+p = re.compile('[A-Za-z ]+$')
```
jgengo, line 45: I wouldn't use regex for a so simple check that can be done with .isalpha()

chall01/mtuomine.py
```diff
@@ -0,0 +1,51 @@
...
+
+if len(sys.argv) != 2 or not p.match(sys.argv[1]):
+	print("usage: ./mtuomine.py <a-zA-Z string>")
```
jgengo, line 48: would be better to use sys.argv[0] than hardcode your login.

chall01/mtuomine.py
```diff
@@ -0,0 +1,51 @@
...
+	for c in msg:
+		if c == ' ':
+			morse += ' '
+		else:
+			morse += MORSES[c]
+	return morse
+
+p = re.compile('[A-Za-z ]+$')
+
+if len(sys.argv) != 2 or not p.match(sys.argv[1]):
```
jgengo, line 47:
```python
if len(sys.argv) !=2 or not sys.argv[1].replace(" ", "").isalpha():
```

chall01/mtuomine.py
```diff
@@ -0,0 +1,51 @@
...
+	print("usage: ./mtuomine.py <a-zA-Z string>")
+	exit()
+
+print(encrypt(sys.argv[1]))
```
jgengo, line 51: A good practice is to protect your python scripts that are meant to be run from running on import.
It's done by
```python
if __name__ == "__main__":
```

jgengo (commented):
really clean code

PR #56, opened 2020-03-26
srouhe morse

Rouhija:
Simple solution:
    1. Check that argc == 2
    2. Check length of argv[1]
    3. lower() it and map if with the morse dictionary (in case of illegal characters, KeyError and display 
        usage
    4. Return encoded morse and print it out

### DIFF
--- a/chall01/srouhe.py
+++ b/chall01/srouhe.py
@@ -1,0 +1,57 @@
+#!/usr/bin/python3
+
+import sys
+
+MAPPING = {
+	'a': '.-',
+	'b': '-...',
+	'c': '-.-.',
+	'd': '-..',
+	'e': '.',
+	'f': '..-.', 
+	'g': '--.', 
+	'h': '....', 
+	'i': '..', 
+	'j': '.---', 
+	'k': '-.-', 
+	'l': '.-..', 
+	'm': '--', 
+	'n': '-.', 
+	'o': '---', 
+	'p': '.--.', 
+	'q': '--.-', 
+	'r': '.-.', 
+	's': '...', 
+	't': '-', 
+	'u': '..-', 
+	'v': '...-', 
+	'w': '.--', 
+	'x': '-..-', 
+	'y': '-.--', 
+	'z': '--..',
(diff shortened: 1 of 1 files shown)


chall01/srouhe.py
```diff
@@ -0,0 +1,57 @@
...
+	'v': '...-', 
+	'w': '.--', 
+	'x': '-..-', 
+	'y': '-.--', 
+	'z': '--..',
+	' ': ' '
+}
+
+def print_usage():
+	print('usage: ./srouhe.py <a-zA-Z string>')
```
jgengo, line 36: would be better to use sys.argv[0] not to hardcode srouhe.py.

chall01/srouhe.py
```diff
@@ -0,0 +1,57 @@
...
+	sys.exit(1)
+
+def encode_arg(s):
+	r = ''
+	for c in s:
+		try:
+			r += MAPPING[c]
+		except KeyError:
+			print_usage()
```
jgengo, line 45: you could avoid try / except for this simple script by making sure you handle all the error case possible (there is not so many)

chall01/srouhe.py
```diff
@@ -0,0 +1,57 @@
...
+
+def main():
+	if len(sys.argv) != 2:
+		print_usage()
+	if len(sys.argv[1]) == 0:
+		print_usage()
+	morse = encode_arg(sys.argv[1].lower())
+	print(morse)
+
+if __name__ == '__main__':
```
jgengo, line 56: really good!

chall01/srouhe.py
```diff
@@ -0,0 +1,57 @@
...
+		try:
+			r += MAPPING[c]
+		except KeyError:
+			print_usage()
+	return r
+
+def main():
+	if len(sys.argv) != 2:
+		print_usage()
+	if len(sys.argv[1]) == 0:
```
jgengo, line 51: `if not sys.argv[1]`

jgengo (commented):
really clean code, nice job

PR #57, opened 2020-03-26
jhakala morse

juhakala:
1. Check for invalid number of arguments and if first argument is empty
2. Loop through argument and and check if letter is space or a-zA-Z character, if not print usage
3.Pick matching character and take Morse code from MORSE dictionary and add to output
4.After looped the whole message print output


chall01/jhakala.py
```diff
@@ -0,0 +1,38 @@
...
+            output += MORSE[letter]
+        else:
+            print("usage: ./jhakala.py <a-zA-Z string>")
+            return
+    print(output)
+    return
+
+def main():
+    if len(sys.argv) != 2 or len(sys.argv[1]) == 0:
+        print("usage: ./jhakala.py <a-zA-Z string>")
```
jgengo, line 33: Would be better to use sys.argv[0] than hardcode your login

chall01/jhakala.py
```diff
@@ -0,0 +1,38 @@
...
+        elif letter >= 'A' and letter <= 'Z':
+            output += MORSE[letter]
+        else:
+            print("usage: ./jhakala.py <a-zA-Z string>")
+            return
+    print(output)
+    return
+
+def main():
+    if len(sys.argv) != 2 or len(sys.argv[1]) == 0:
```
jgengo, line 32:
```python
or not sys.argv[1]
```

chall01/jhakala.py
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env python3
+import sys
+
+MORSE = { 'A':'.-', 'B':'-...', 'C':'-.-.', 'D':'-..', 'E':'.',
```
jgengo, line 4: could have just added lowercase ones.

chall01/jhakala.py
```diff
@@ -0,0 +1,38 @@
...
+
+def tomorse(message): 
+    output = "" 
+    for letter in message: 
+        if letter == ' ':
+            output += " "
+        elif letter >= 'a' and letter <= 'z':
+            output += MORSE[letter]
+        elif letter >= 'A' and letter <= 'Z':
+            output += MORSE[letter]
```
jgengo, line 24: could have avoided that with 
```python
elif letter.isalpha():
  output += MORSE[letter.lower()]
```

chall01/jhakala.py
```diff
@@ -0,0 +1,38 @@
...
+    print(output)
+    return
+
+def main():
+    if len(sys.argv) != 2 or len(sys.argv[1]) == 0:
+        print("usage: ./jhakala.py <a-zA-Z string>")
+        return
+    tomorse(sys.argv[1])
+
+if __name__ == '__main__':
```
jgengo, line 37: Really good!

PR #58, opened 2020-03-26
Chall01 morsecode by sluhtala

sluhtala:
Hello,

First I imported sys to get access to arguments. Then check if there are only 1 argument. Then I go through all the letters in argument and check if they are included in my alphabet variable. If they are all included I create a list of strings for the morse code. And now it's just going through the list again printing morse code based on the ascii order. Line break in the end.


chall01/sluhtala.py
```diff
@@ -0,0 +1,30 @@
+import sys
+
+def print_usage():
+    print("usage: ./sluhtala.py <a-zA-Z string>")
```
jgengo, line 4: you could use sys.argv[0] not to hardcode your script name into the usage.

chall01/sluhtala.py
```diff
@@ -0,0 +1,30 @@
+import sys
+
+def print_usage():
```
jgengo, line 3: Good! you applied the DRY (don't repeat yourself) concept with this function!

chall01/sluhtala.py
```diff
@@ -0,0 +1,30 @@
+import sys
```
jgengo, line 1: You forgot the shebang to execute the script 

https://stackoverflow.com/questions/7670303/purpose-of-usr-bin-python3/7670338

chall01/sluhtala.py
```diff
@@ -0,0 +1,30 @@
...
+def print_usage():
+    print("usage: ./sluhtala.py <a-zA-Z string>")
+    exit()
+
+if len(sys.argv) != 2:
+    print_usage()
+arglen = len(sys.argv[1])
+if arglen == 0:
+    print_usage()
+alphabet = 'abcdefghijklmnopqrstuvxyzABCDEFGHIJKLMNOPQRSTUVXYZ '
```
jgengo, line 12:
```python
alphabet = string.ascii_lowercase
```

chall01/sluhtala.py
```diff
@@ -0,0 +1,30 @@
...
+
+def print_usage():
+    print("usage: ./sluhtala.py <a-zA-Z string>")
+    exit()
+
+if len(sys.argv) != 2:
+    print_usage()
+arglen = len(sys.argv[1])
+if arglen == 0:
+    print_usage()
```
jgengo, line 11:
```python
if not sys.argv[1]:
  print_usage()
```

chall01/sluhtala.py
```diff
@@ -0,0 +1,30 @@
...
+alphabet = 'abcdefghijklmnopqrstuvxyzABCDEFGHIJKLMNOPQRSTUVXYZ '
+for i in range(arglen):
+    if sys.argv[1][i]  not in alphabet:
+        print_usage()
```
jgengo, line 15: you can use isalpha()

chall01/sluhtala.py
```diff
@@ -0,0 +1,30 @@
...
+morse = ['.-','-...','-.-.',
+'-..','.','..-.','--.','....',
+'..','.---','-.-','.-..','--',
+'-.','---','.--.','--.-','.-.',
+'...','-','..-','...-','.--','-..-',
+'-.--','--..']
+for i in range(arglen):
```
jgengo, line 22:
```python
for c in sys.argv[1]:
```
so it will loop on every character.

chall01/sluhtala.py
```diff
@@ -0,0 +1,30 @@
...
+    letter = sys.argv[1][i]
+    if letter.isupper():
+        letter = letter.lower()
+    if (sys.argv[1][i] == ' '):
+        print(' ', end='')
+    else:
+        print(morse[ord(letter) - 97], end='')
+print('')
```
jgengo, line 30: A good practice is to protect your python scripts that are meant to be run from running on import.
It's done by
```python
if __name__ == "__main__":
```

chall01/sluhtala.py
```diff
@@ -0,0 +1,30 @@
...
+'..','.---','-.-','.-..','--',
+'-.','---','.--.','--.-','.-.',
+'...','-','..-','...-','.--','-..-',
+'-.--','--..']
+for i in range(arglen):
+    letter = sys.argv[1][i]
+    if letter.isupper():
+        letter = letter.lower()
+    if (sys.argv[1][i] == ' '):
+        print(' ', end='')
```
jgengo, line 27: Would be better to append in a string and then print the string, to avoid printing char by char.

jgengo:
Unfortunately, there is no shebang, I have to close this PR.

PR #59, opened 2020-03-26
llahti chall01

lapaset:
I solved the problem by creating a list with all the morse alpabet in it. Then i check if the input is valid, first by checking the amount of arguments, then if the argument is an empty string and finally by looping each character of the string. If not valid, usage is printed. If yes, i check if the character is space, uppercase or lowercase. Space is added to the string to be printed as it is. Characters are turned into corresponding morse by counting the index by their ascii values.

### DIFF
--- a/chall01/llahti.py
+++ b/chall01/llahti.py
@@ -1,0 +1,48 @@
+#!/usr/bin/env python3
+import sys
+
+morse = [".-",
+"-...",
+"-.-.",
+"-..",
+".",
+"..-.",
+"--.",
+"....",
+"..", 
+".---",
+"-.-",
+".-..",
+"--",
+"-.",
+"---",
+".--.",
+"--.-",
+".-.",
+"...",
+"-",
+"..-",
+"...-",
+".--",
+"-..-",
+"-.--",
+"--.."]
+
+if len(sys.argv) != 2 or sys.argv[1] == "":
+    print("usage: ./xlogin.py <a-zA-Z string>")
+    exit()
+
+string = ""
+
+for c in sys.argv[1]:
+    if c.isalpha() == False and c != ' ':
+        print("usage: ./xlogin.py <a-zA-Z string>")
+        exit()
+    if c == ' ':
+        string += ' '
+    elif c.islower():
+        string += morse[ord(c) - ord('a')]
+    else:
+        string += morse[ord(c) - ord('A')]
+
+print(string)


chall01/llahti.py
```diff
@@ -0,0 +1,48 @@
...
+
+morse = [".-",
+"-...",
+"-.-.",
+"-..",
+".",
+"..-.",
+"--.",
+"....",
+"..",
```
jgengo, line 12: like this array is not really describeful like could be a Dict, I would recommend put it in a one-line.

chall01/llahti.py
```diff
@@ -0,0 +1,48 @@
...
+    if c.isalpha() == False and c != ' ':
+        print("usage: ./xlogin.py <a-zA-Z string>")
+        exit()
+    if c == ' ':
+        string += ' '
+    elif c.islower():
+        string += morse[ord(c) - ord('a')]
+    else:
+        string += morse[ord(c) - ord('A')]
+
```
jgengo, line 47: A good practice is to protect your python scripts that are meant to be run from running on import.
It's done by
```python
if __name__ == "__main__":
```

chall01/llahti.py
```diff
@@ -0,0 +1,48 @@
...
+"-",
+"..-",
+"...-",
+".--",
+"-..-",
+"-.--",
+"--.."]
+
+if len(sys.argv) != 2 or sys.argv[1] == "":
+    print("usage: ./xlogin.py <a-zA-Z string>")
```
jgengo, line 32: you were supposed to write your login.py not `llahti.py`

you could have retrieved this information with `sys.argv[0]`

chall01/llahti.py
```diff
@@ -0,0 +1,48 @@
...
+"...",
+"-",
+"..-",
+"...-",
+".--",
+"-..-",
+"-.--",
+"--.."]
+
+if len(sys.argv) != 2 or sys.argv[1] == "":
```
jgengo, line 31:
```python
if len(sys.argv) != 2 or not sys.argv[1]
```

chall01/llahti.py
```diff
@@ -0,0 +1,48 @@
...
+    print("usage: ./xlogin.py <a-zA-Z string>")
+    exit()
+
+string = ""
+
+for c in sys.argv[1]:
+    if c.isalpha() == False and c != ' ':
+        print("usage: ./xlogin.py <a-zA-Z string>")
+        exit()
```
jgengo, line 40: isalpha() can be used on string and char.

So I would suggest:

```python
if not sys.argv[1].replace(" ", "").isalpha():
  print...
  exit...
```

chall01/llahti.py
```diff
@@ -0,0 +1,48 @@
...
+    if c == ' ':
+        string += ' '
+    elif c.islower():
+        string += morse[ord(c) - ord('a')]
+    else:
+        string += morse[ord(c) - ord('A')]
```
jgengo, line 46: like we already checked it is alpha, you can just go with:

```python
string += morse[ord(c.lower()) - ord('a')]
```

PR #60, opened 2020-03-26
final version, added extra protection for scandics

MikeyLHolm:
Basic idea is to go through the input_str characters one by one and find the corresponding morse_code from the created array. Obviously array starts from 0 and smallest letter from 1 thus the "letter - 1". I found it more readable than subtracting the 1 already from y.

There is specific error() function for error messages since its used more than once. Everything but " ", A-Z and a-z provide this error.

This is created in Windows env so I can't be sure that it works on other systems (the #!/... at start). However I provided the usage: ./..py even though on PC i use the following:

`python mlindhol.py "daily sos" `

I wish I could have done the scandics (öäå etc) with isalpha but ran out of skills and had to do it with a long if. My original version was something like:
```
if any(x.isalpha() is False and x.isspace() is False for x in input_str)
    error()
```

### DIFF
--- a/chall01/mlindhol.py
+++ b/chall01/mlindhol.py
@@ -1,0 +1,29 @@
+#!/usr/bin/env python
+import sys
+
+
+def error():
+    print("usage: ./mlindhol.py <a-zA-Z string>")
+    exit(-1)
+
+
+if len(sys.argv) != 2 or len(sys.argv[1]) == 0:
+    error()
+
+morse_array = [".-", "-...", "-.-.", "-..", ".", "..-.", "--.", "....", "..", ".---", "-.-", ".-..", "--",
+               "-.", "---", ".--.", "--.-", ".-.", "...", "-", "..-", "...-", ".--", "-..-", "-.--", "--.."]
+
+input_str = sys.argv[1]
+
+for x in input_str:
+    av = ord(x)
+    if av < 32 or 32 < av < 65 or 90 < av < 97 or av > 122:
+        error()
+
+for s in input_str:
+    if s is " ":
+        print(" ", end="")
+    else:
+        y = 64 if s.isupper() else 96
+        letter = ord(s) - y
+        print(morse_array[letter - 1], end="")


chall01/mlindhol.py
```diff
@@ -0,0 +1,29 @@
...
+
+morse_array = [".-", "-...", "-.-.", "-..", ".", "..-.", "--.", "....", "..", ".---", "-.-", ".-..", "--",
+               "-.", "---", ".--.", "--.-", ".-.", "...", "-", "..-", "...-", ".--", "-..-", "-.--", "--.."]
+
+input_str = sys.argv[1]
+
+for x in input_str:
+    av = ord(x)
+    if av < 32 or 32 < av < 65 or 90 < av < 97 or av > 122:
+        error()
```
jgengo, line 21:
```python
if not x.isalpha():
  error()
```
more pythonic way to do that.

chall01/mlindhol.py
```diff
@@ -0,0 +1,29 @@
...
+
+if len(sys.argv) != 2 or len(sys.argv[1]) == 0:
+    error()
+
+morse_array = [".-", "-...", "-.-.", "-..", ".", "..-.", "--.", "....", "..", ".---", "-.-", ".-..", "--",
+               "-.", "---", ".--.", "--.-", ".-.", "...", "-", "..-", "...-", ".--", "-..-", "-.--", "--.."]
+
+input_str = sys.argv[1]
+
+for x in input_str:
```
jgengo, line 18: you can actually even avoid it by doing :

```python
if not input_str.replace(" ", "").isalpha():
```

chall01/mlindhol.py
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env python
+import sys
+
+
+def error():
+    print("usage: ./mlindhol.py <a-zA-Z string>")
+    exit(-1)
+
+
+if len(sys.argv) != 2 or len(sys.argv[1]) == 0:
```
jgengo, line 10:
```python
or not sys.argv[1]
```

chall01/mlindhol.py
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env python
+import sys
+
+
+def error():
+    print("usage: ./mlindhol.py <a-zA-Z string>")
```
jgengo, line 6: would be better to use sys.argv[0] than hardcode your login.

chall01/mlindhol.py
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env python
```
jgengo, line 1: python3

chall01/mlindhol.py
```diff
@@ -0,0 +1,29 @@
...
+    av = ord(x)
+    if av < 32 or 32 < av < 65 or 90 < av < 97 or av > 122:
+        error()
+
+for s in input_str:
+    if s is " ":
+        print(" ", end="")
+    else:
+        y = 64 if s.isupper() else 96
+        letter = ord(s) - y
```
jgengo, line 28: smart :) 

or
```python
letter = ord(s.lower()) - 96
```

chall01/mlindhol.py
```diff
@@ -0,0 +1,29 @@
...
+        print(morse_array[letter - 1], end="")
```
jgengo, line 29: would be better avoiding printing char by char but to append into a string and then print the string at the end.

and

It's done by
```python
if __name__ == "__main__":
```

chall01/mlindhol.py
MikeyLHolm: isalpha() accepts ä, ö and å (and other scandic letters i think) but there is no equivalent morse code for those that is why I change it to this if-madness. But that is nice short way of doing it. This was my original way of doing it: 
```
if any(x.isalpha() is False and x.isspace() is False for x in input_str)
    error()
```

PR #61, opened 2020-03-26
PR chall01

shilena91:
My logic:
- I create an array of morses
- Make standard input string to lowercase
- Check input string, only accept if it contains alphabet and space, otherwise print usage and exit
- Create morse_string dictionary with keys are alphabet A-Z and values are morses respectively
- Iterate the input string, take its characters, print the morse value that has the key matches with input string's characters.

I tried to fetch upstream from master and merge. Sorry if I messed up git's log.


chall01/hopham.py
```diff
@@ -0,0 +1,31 @@
...
+
+import sys
+import string
+
+morse = [".-", "-...", "-.-.", "-..", ".", "..-.", "--.", "....", "..", 
+        ".---", "-.-", ".-..", "--", "-.", "---", ".--.", "--.-", ".-.", 
+        "...", "-", "..-", "...-", ".--", "-..-", "-.--", "--.."]
+
+if (len(sys.argv) != 2 or len(sys.argv[1]) == 0):
+    print("usage: ./hopham.py <a-zA-Z string>")
```
jgengo, line 11: don't hardcode the script name

```python
```

chall01/hopham.py
```diff
@@ -0,0 +1,31 @@
...
+    exit()
+argument = sys.argv[1]
+argument_lowcase = argument.lower()
+for character in argument_lowcase:
+    if (character not in string.ascii_lowercase and character != " "):
+        print("usage: ./hopham.py <a-zA-Z string>")
+        exit()
```
jgengo, line 18: can be done with
```python
if not argument.replace(" ", "").isalpha():
```

chall01/hopham.py
```diff
@@ -0,0 +1,31 @@
...
+
+morse_string = dict()
+i = 0
+for key in string.ascii_lowercase:
+    morse_string[key] = morse[i]
+    i += 1
```
jgengo, line 24: the pythonic way to do the same thing is:

```python
morse_string = {k: v for k, v in zip(string.ascii_lowercase, morse)}
```

Using a dict was a really good idea!

chall01/hopham.py
```diff
@@ -0,0 +1,31 @@
...
+
+for i in argument_lowcase:
+    if (i in morse_string):
+        sys.stdout.write(morse_string[i])
+    else:
+        sys.stdout.write(i)
```
jgengo, line 30: like in C, try to use write only one time and not for every char.

you could have appended to a string every char and at the end print the string.

chall01/hopham.py
```diff
@@ -0,0 +1,31 @@
...
+sys.stdout.write("\n")
```
jgengo, line 31: just a good practice: add a new line at the end of your file

jgengo:
Sadly, I've to close your PR; you pushed again all the C files.

Don't forget to git pull before working and doing your PR. 
And never use `git push -f`

PR #62, opened 2020-03-26
morse

Rakutaa:
maybe this is the pr or maybe not, anyhow..

first made dictionary, 
then check that there will be only one argument for program otherwise: usage,
then making this arg as a lowercase and passing it to encoding function,
then in a function moving char by char the given arg and adding dictionary value to a new empty string,
if there is no key for the char, will print usage and exit. 
otherwise will print the new string.


chall01/vtran.py
```diff
@@ -0,0 +1,53 @@
...
+def usage():
+    print ('usage: ./vtran.py <a-zA-Z string>')
+    sys.exit(1)
+
+def main():
+    if len (sys.argv) == 2:
+        tinnitus(sys.argv[1].lower())
+    else:
+        usage()
+main()
```
jgengo, line 53: A good practice is to protect your python scripts that are meant to be run from running on import.
It's done by:
```python
if __name__ == "__main__":
```

chall01/vtran.py
```diff
@@ -0,0 +1,53 @@
...
+    if ret:
+        print (ret)
+
+def usage():
+    print ('usage: ./vtran.py <a-zA-Z string>')
+    sys.exit(1)
+
+def main():
+    if len (sys.argv) == 2:
+        tinnitus(sys.argv[1].lower())
```
jgengo, line 50: What tinnitus means? You should use meaningful function names.

chall01/vtran.py
```diff
@@ -0,0 +1,53 @@
...
+    's': "...",
+    't': "-",
+    'u': "..-",
+    'v': "...-",
+    'w': ".--",
+    'x': "-..-",
+    'y': "-.--",
+    'z': "--..",
+    ' ': " "
+}
```
jgengo, line 32: Would have been better to write it in one-line and use a meaningful variable name

chall01/vtran.py
```diff
@@ -0,0 +1,53 @@
...
+
+def tinnitus(str):
+    ret = ""
+    for i in str:
+        try:
+            ret += tittut[i]
+        except KeyError:
+            usage()
```
jgengo, line 40: I think for this script using try-except isn't necessary and would have been better you try to handle the input error before letting your code raise an issue and catch it to display usage.

jgengo (approved):
Be careful to use meaningful variable and function name.

It's good you know how to use try-catch/except but I would suggest avoiding using it if you can handle error case by yourself.

PR #64, opened 2020-03-26
chall01 ismelich

iljaSL:
My solution for the chall01 challenge.

### DIFF
--- a/chall01/ismelich.py
+++ b/chall01/ismelich.py
@@ -30,16 +30,18 @@ morse_code = {
     'Y':'-.--',
     'Z':'--..',
 }
+
+
+if len(sys.argv) != 2:
+    print ('usage: ./ismelich.py <a-zA-Z string>')
+    sys.exit(1)
+
 str = sys.argv[1]
 for x in str:
     av = ord(x)
     if av < 32 or 32 < av < 65 or 90 < av < 97 or av > 122:
         print('usage: ./ismelich.py <a-zA-Z string>')
-        exit(1)
-
-if len(sys.argv) != 2:
-    print ('usage: ./ismelich.py <a-zA-Z string>')
-    sys.exit(1)
+        sys.exit(1)
 
 def encryptor(text):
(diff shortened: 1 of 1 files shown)


chall01/ismelich.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+if len(sys.argv) != 2:
+    print ('usage: ./ismelich.py <a-zA-Z string>')
+    sys.exit(1)
+
+str = sys.argv[1]
+for x in str:
+    av = ord(x)
+    if av < 32 or 32 < av < 65 or 90 < av < 97 or av > 122:
```
jgengo, line 42: you can use isalpha() and replace() to avoid looping on every char of the string and using ascii value to check
```python
if not sys.argv[1].replace(" ", "").isalpha()
```

chall01/ismelich.py
```diff
@@ -0,0 +1,57 @@
...
+        print('usage: ./ismelich.py <a-zA-Z string>')
```
jgengo, line 43: avoid using hardcoded string.
```python
```

chall01/ismelich.py
```diff
@@ -0,0 +1,57 @@
...
+    encrypted_text = ""
+    for letters in text:
+        if letters != " ":
+            encrypted_text = encrypted_text + morse_code.get(letters)
+        else:
+            encrypted_text += " "
+    print(encrypted_text)
+
+
+text_to_encrypt = " ".join(sys.argv[1:]).upper()
```
jgengo, line 56: Regarding the example given into the README.md you should print the usage `if len(sys.argv) != 2` ; but really clean way to write it! Good usage of join and upper()

chall01/ismelich.py
```diff
@@ -0,0 +1,57 @@
...
+encryptor(text_to_encrypt)
```
jgengo, line 57: don't forget a new line at the end, it's a good practice.

PR #65, opened 2020-03-26
Chall01

mdesta:
Preliminaries
A function for mapping characters into morse code
A function to printout USAGE message
A function to check if the string to be encodes contains only alpha or Space and No accented
characters like ä or ö
=> Uses isalpha() isspace() AND 'not in' operator
A function to encode a valid string into morse code
=> Since Mapping function is defined only for small cases, first change string to lower case
=> For every char in the string (NOT Space), call mapping function and concatenate to
output string
=> If space, simply concatenate to output string
=> Return the output string
Main Function Uses sys to get access to argument passed, and decide with helper function to call depending on the argument type(s) passed


chall01/mdesta.py
```diff
@@ -0,0 +1,107 @@
...
+	elif x is 'b': 
+		return "-..."
+	elif x is 'c': 
+		return "-.-."
+	elif x is 'd': 
+		return "-.."
+	elif x is 'e': 
+		return "."
+	elif x is 'f': 
+		return "..-."
```
jgengo, line 16: If-forest, it's not really pythonic, and this should be avoided in all the language

You should have used Python Dict : https://docs.python-guide.org/writing/style/

```python
mydict = {'a': '.-', 'b': '-...', 'c': '-.-.', 'd': '-..', 'e': '.', 'f': '..-.', 'g': '--.', 'h': '....', 'i': '..', 'j': '.---', 'k': '-.-', 'l': '.-..', 'm': '--', 'n': '-.', 'o': '---', 'p': '.--.', 'q': '--.-', 'r': '.-.', 's': '...', 't': '-', 'u': '..-', 'v': '...-', 'w': '.--', 'x': '-..-', 'y': '-.--', 'z': '--..'}
```

Then `return mydict[x]`

chall01/mdesta.py
```diff
@@ -0,0 +1,107 @@
...
+		return "-.--"
+	elif x is 'z': 
+		return "--.."
+
+# Encode one Char at a time
+# concatnate to a string 
+# Return encoded string
+def morseEncode(string): 
+    encodedString = ""
+    string = string.lower()
```
jgengo, line 63: Yes! Smart!

chall01/mdesta.py
```diff
@@ -0,0 +1,107 @@
...
+    for char in string:
```
jgengo, line 64: can be used here: 
```python
 for char in string.lower()
```

chall01/mdesta.py
```diff
@@ -0,0 +1,107 @@
...
+        if char == " ":
+            encodedString += " "
+        else:
+            encodedString += encodeToMorse(char)
+    return encodedString
+
+# Prints Usage 
+def printUsage():
+    print("usage: ./mdesta.py <a-zA-Z string>")
```
jgengo, line 73: better not to hardcode your app name but use `sys.argv[0]`

chall01/mdesta.py
```diff
@@ -0,0 +1,107 @@
...
+# Start Encoding if
+# -> ONLY one argument passed
+# -> AND it is NOT empty
+# -> AND does NOT contain accented characters
+# -> Print the returned encoded string
+# Otherwise print USAGE
+import sys
+if len(sys.argv) == 2:
+    argument = sys.argv[1]
+    if len(argument) != 0 and isAlphaOrSpace(argument):
```
jgengo, line 101:
```python
if len(argument) != 0 and argument.replace(" ", "").isalpha():
```

jgengo (commented):
Always If-forest in every language, with such a high-level language you can be sure almost every function already exists to do so simple things like isalpha() or the dict 'letter' -> 'morse'

I suggest you read: https://docs.python-guide.org/writing/style/

PR #66, opened 2020-03-26
Chall01 Morse

Epicurius:
Checks how many arguments.
Cycles through every character and checks if they matches with "abcdefghijklmnopqrstuvwxyz".
Checks with cheat_sheet for specific morse code or space.
If there is morse code than prints else error message.

### DIFF
--- a/chall01/nneronin.py
+++ b/chall01/nneronin.py
@@ -1,0 +1,35 @@
+#!/usr/bin/env python3
+import sys
+
+CHEAT_SHEET = {'a':'.-', 'b':'-...', 
+               'c':'-.-.', 'd':'-..',
+               'e':'.', 'f':'..-.',
+               'g':'--.', 'h':'....', 
+               'i':'..', 'j':'.---',
+               'k':'-.-', 'l':'.-..',
+               'm':'--', 'n':'-.',
+               'o':'---', 'p':'.--.',
+               'q':'--.-', 'r':'.-.',
+               's':'...', 't':'-',
+               'u':'..-', 'v':'...-',
+               'w':'.--', 'x':'-..-',
+               'y':'-.--', 'z':'--..'}
+
+def error_message():
+    print ("usage:", sys.argv[0], "<a-zA-Z string>")
+    quit()
+
+if len(sys.argv) != 2:
+    error_message()
+morse = ''
+alphabet = "abcdefghijklmnopqrstuvwxyz"
+for letter in sys.argv[1]:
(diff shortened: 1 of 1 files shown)


chall01/nneronin.py
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env python3
+import sys
+
+CHEAT_SHEET = {'a':'.-', 'b':'-...',
```
jgengo, line 4: variable name could be better charToMorse = {}

chall01/nneronin.py
```diff
@@ -0,0 +1,35 @@
...
+               'o':'---', 'p':'.--.',
+               'q':'--.-', 'r':'.-.',
+               's':'...', 't':'-',
+               'u':'..-', 'v':'...-',
+               'w':'.--', 'x':'-..-',
+               'y':'-.--', 'z':'--..'}
+
+def error_message():
+    print ("usage:", sys.argv[0], "<a-zA-Z string>")
+    quit()
```
jgengo, line 20: good!

chall01/nneronin.py
```diff
@@ -0,0 +1,35 @@
...
+
+if len(sys.argv) != 2:
+    error_message()
+morse = ''
+alphabet = "abcdefghijklmnopqrstuvwxyz"
```
jgengo, line 25: `alphabet = string.ascii_lowercase`

chall01/nneronin.py
```diff
@@ -0,0 +1,35 @@
...
+for letter in sys.argv[1]:
+    if alphabet.find(letter.lower()) == -1 and letter != ' ':
```
jgengo, line 27: clearer
```python
if not letter.isalpha() and letter != ' ':
```

chall01/nneronin.py
```diff
@@ -0,0 +1,35 @@
...
+        error_message()
+    if letter != ' ':
+        morse += CHEAT_SHEET[letter.lower()]
```
jgengo, line 30: could avoid that by adding space into your dict

chall01/nneronin.py
```diff
@@ -0,0 +1,35 @@
...
+    else:
+        morse += ' '
+if morse == '':
+    error_message()
+print (morse)
```
jgengo, line 35: A good practice is to protect your python scripts that are meant to be run from running on import.
It's done by
```python
if __name__ == "__main__":
```

PR #67, opened 2020-03-26
spentti morse code

saraarmiida:
I created a dictionary to be able to translate letters to morse. I check if the input is valid, then i index through the word that was sent as parameter and print out the corresponding morse code for each letter. I added space to the dictionary to be able to print sentences.

### DIFF
--- a/chall01/spentti.py
+++ b/chall01/spentti.py
@@ -1,0 +1,50 @@
+#!/usr/bin/env python3
+import sys
+import re
+
+morse = {
+	'a': '.-',
+	'b': '-...',
+	'c': '-.-.',
+	'd': '-..',
+	'e': '.',
+	'f': '..-.',
+	'g': '--.',
+	'h': '....',
+	'i': '..',
+	'j': '.---',
+	'k': '-.-',
+	'l': '.-..',
+	'm': '--',
+	'n': '-.',
+	'o': '---',
+	'p': '.--.',
+	'q': '--.-',
+	'r': '.-.',
+	's': '...',
+	't': '-',
+	'u': '..-',
+	'v': '...-',
+	'w': '.--',
+	'x': '-..-',
+	'y': '-.--',
+	'z': '--..',
+	' ': ' '
+}
+
+if (len(sys.argv) == 2):
+	word = sys.argv[1].lower().strip()
+	mode = 0
+	for c in word:
+		if c.isalpha() or c == ' ':
+			mode = mode
+		else:
+			mode = 1
+	if (mode == 0 and len(word) > 0):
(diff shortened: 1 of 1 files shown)


chall01/spentti.py
```diff
@@ -0,0 +1,50 @@
...
+	'u': '..-',
+	'v': '...-',
+	'w': '.--',
+	'x': '-..-',
+	'y': '-.--',
+	'z': '--..',
+	' ': ' '
+}
+
+if (len(sys.argv) == 2):
```
jgengo, line 35: to avoid your code to be too much indented; I would recommend going the other way
```python
if len(sys.argv) != 2:
  print_usage()
  exit(1)

word = sys.argv[1].lower.strip()
```

chall01/spentti.py
```diff
@@ -0,0 +1,50 @@
...
+	word = sys.argv[1].lower().strip()
+	mode = 0
+	for c in word:
+		if c.isalpha() or c == ' ':
+			mode = mode
+		else:
+			mode = 1
```
jgengo, line 42:
```python
word = sys.argv[1].lower()
mode =  1 if word.replace(" ", "").isalpha() else 0
```

chall01/spentti.py
```diff
@@ -0,0 +1,50 @@
...
+	if (mode == 0 and len(word) > 0):
```
jgengo, line 43:
```python
word = sys.argv[1].lower()
if word and word.replace(" ", "").isalpha():
```

with this, you don't even need what is just above

chall01/spentti.py
```diff
@@ -0,0 +1,50 @@
...
+		for char in word:
+			sys.stdout.write(morse[char])
```
jgengo, line 45: like in C, try to avoid as much as possible to use syscall. So you shouldn't print char by char but append morse[char] to a string and then print this string at the end.

chall01/spentti.py
```diff
@@ -0,0 +1,50 @@
...
+		print('')
+	else:
+		print("usage: ./spentti.py <a-zA-Z string>")
```
jgengo, line 48: you should also avoid to hardcode information that could change like the script name.

`sys.argv[0]` is where your first argument (the script name) is stored.

chall01/spentti.py
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env python3
+import sys
+import re
```
jgengo, line 3: I see `re` but can't see any regex in your code. Make sure not to import a library you are not using.

jgengo (approved):

PR #68, opened 2020-03-26
Chall01 - solution by hhuhtane

huhtasheikki:
### hhuhtane solution to Chall01

**Variables:**
- **alphabet** = lower case letters
- **capital** = upper case letters
- **arg_i** = index for letters of arg[1]
- **codes** = array of strings, single string containing each morse code for specific letter
- **i** = index for alphabet, capital and codes variable
- **morse** = empty string to store all the morsed letters

**What program does and why:**

1. Check that user has inserted only one argument and it contains only letters
-> if not print the usage
2. Analyze every character of inserted argument and transform the letters to morse code
-> store that data to the string called morse
3. After whole inserted argument has been morsed, print it


chall01/hhuhtane.py
```diff
@@ -0,0 +1,37 @@
...
+#                                                 +#+#+#+#+#+   +#+            #
+#    Created: 2020/03/26 15:47:35 by hhuhtane          #+#    #+#              #
+#    Updated: 2020/03/26 16:51:09 by hhuhtane         ###   ########.fr        #
+#                                                                              #
+# **************************************************************************** #
+
+import sys
+
+alphabet = "abcdefghijklmnopqrstuvwxyz"
+capital = "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
```
jgengo, line 18: you can use python built-in function for that:
```python
alphabet = string.ascii_lowercase
```

you did not need to have a capital array

chall01/hhuhtane.py
```diff
@@ -0,0 +1,37 @@
...
+arg_i = 0
+i = 0
+morse = ""
+codes = ['.-', '-...', '-.-.', '-..', '.', '..-.', '--.', '....', '..', '.---',
+         '-.-', '.-..', '--', '-.', '---', '.--.', '--.-', '.-.', '...', '-',
+         '..-', '...-', '.--', '-..-', '-.--', '--..']
+
+if len(sys.argv) == 2:
+    while arg_i < len(sys.argv[1]):
```
jgengo, line 29: I would suggest using python `for`:

```python
for char in sys.argv[1]:
```

You could have removed arg_i index

chall01/hhuhtane.py
```diff
@@ -0,0 +1,37 @@
...
+        i = 0
+        while i < 26:
+            if alphabet[i] == sys.argv[1][arg_i]:
+                morse += codes[i]
+            if capital[i] == sys.argv[1][arg_i]:
+                morse += codes[i]
```
jgengo, line 35: I need to check your code but I don't really see where you check there isn't another non alpha character.

You could have create a Python Dictionary : https://www.tutorialspoint.com/python3/python_dictionary.htm

```python
{'a': '.-', 'b': '-...', 'c': '-.-.', 'd': '-..', 'e': '.', 'f': '..-.', 'g': '--.', 'h': '....', 'i': '..', 'j': '.---', 'k': '-.-', 'l': '.-..', 'm': '--', 'n': '-.', 'o': '---', 'p': '.--.', 'q': '--.-', 'r': '.-.', 's': '...', 't': '-', 'u': '..-', 'v': '...-', 'w': '.--', 'x': '-..-', 'y': '-.--', 'z': '--..'}
```

and then do something like 
```python
morse += mydict[arg_i]
```

chall01/hhuhtane.py
```diff
@@ -0,0 +1,37 @@
...
+#                                                                              #
+#                                                         :::      ::::::::    #
+#    hhuhtane.py                                        :+:      :+:    :+:    #
+#                                                     +:+ +:+         +:+      #
+#    By: hhuhtane <hhuhtane@student.hive.fi>        +#+  +:+       +#+         #
+#                                                 +#+#+#+#+#+   +#+            #
+#    Created: 2020/03/26 15:47:35 by hhuhtane          #+#    #+#              #
+#    Updated: 2020/03/26 16:51:09 by hhuhtane         ###   ########.fr        #
+#                                                                              #
+# **************************************************************************** #
```
jgengo: Don't put the 42 header for those challenges, please

It is missing here the shebang to make your python script executable:

https://stackoverflow.com/questions/7670303/purpose-of-usr-bin-python3/7670338

jgengo (commented):
Sadly you forget the shebang, 

I guess it's your first time with python and I can see the C language influence your way to write code, I would suggest reading: https://docs.python-guide.org/writing/style/

Unfortunately, I will close this PR because you pushed the README.md again (I can't find how and why)

PR #69, opened 2020-03-26
ptuukkan's solution

ptuukkan:
My solution's idea is to check for each character if we have a mapping in our codes dictionary, then concatenate its code in result string.

### DIFF
--- a/chall01/ptuukkan.py
+++ b/chall01/ptuukkan.py
@@ -1,0 +1,45 @@
+#!/usr/bin/python3
+
+import sys
+
+codes = {
+	"a": ".-",
+	"b": "-...",
+	"c": "-.-.",
+	"d": "-..",
+	"e": ".",
+	"f": "..-.",
+	"g": "--.",
+	"h": "....",
+	"i": "..",
+	"j": ".---",
+	"k": "-.-",
+	"l": ".-..",
+	"m": "--",
+	"n": "-.",
+	"o": "---",
+	"p": ".--.",
+	"q": "--.-",
+	"r": ".-.",
+	"s": "...",
+	"t": "-",
+	"u": "..-",
+	"v": "...-",
+	"w": ".--",
+	"x": "-..-",
+	"y": "-.--",
+	"z": "--..",
+	" ": " "
+}
+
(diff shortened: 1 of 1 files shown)


chall01/ptuukkan.py
```diff
@@ -0,0 +1,45 @@
...
+	"s": "...",
+	"t": "-",
+	"u": "..-",
+	"v": "...-",
+	"w": ".--",
+	"x": "-..-",
+	"y": "-.--",
+	"z": "--..",
+	" ": " "
+}
```
d15ky, line 33: using dictionary - smart! 
idk if you hardcoded it but if you did, you could compose it automatically in python like that: 
```
morse_string = '.- -... -.-. -.. . ..-. --. .... .. .--- -.- .-.. -- -. --- .--. --.- .-. ... - ..- ...- .-- -..- -.-- --..'
morse_list = morse_string.split(" ") # '\n' if you take one from the assignment
codes = dict(zip(string.ascii_lowercase, morse_list))
```

chall01/ptuukkan.py
```diff
@@ -0,0 +1,45 @@
...
+
+if (len(sys.argv) != 2 or (len(sys.argv[1]) == 0)):
+	exit("usage: ./ptuukkan.py <a-zA-Z string>")
```
d15ky, line 36: sys.argv[0] would be better than hardcoded name

chall01/ptuukkan.py
```diff
@@ -0,0 +1,45 @@
...
+
+result = ""
+
+for char in sys.argv[1]:
+	if char.lower() in codes:
+		result += codes[char.lower()]
+	else:
+		exit("usage: ./ptuukkan.py <a-zA-Z string>")
```
d15ky, line 44: More "pythonic" way to do that would be: 
```
1) check string for bad characters, e.g.
if not sys.argv[1].lower().replace(" ", "").isalpha()
2) form a string in one line and print it:
print("".join(codes[c] for c in sys.argv[1].lower()))
```

d15ky (approved):
Overall, great job discovering Python! Code is clean and smart. I have some minor suggestions for you to think about but nothing crucial :)

PR #70, opened 2020-03-26
chall01

J0NY97:
checks the requirements of the string passed in to the script.
prints out the corresponding morse code to the chars in the string.
if there is an error show usage.

### DIFF
--- a/chall01/jsalmi.py
+++ b/chall01/jsalmi.py
@@ -29,7 +29,7 @@ alpha = {
         'y' : "-.--",
         'z' : "--..",
         ' ' : " "}
-if len(sys.argv) == 2 and len(sys.argv[1]) > 0 and all(i.isalpha() or i.isspace() for i in sys.argv[1]):
+if len(sys.argv) == 2 and len(sys.argv[1]) > 0 and all(i in alpha for i in sys.argv[1]):
     for letter in sys.argv[1]:
         print(alpha[letter.lower()], end = '')
     print('')
--- a/chall01/jsalmi.py
+++ b/chall01/jsalmi.py
@@ -1,0 +1,37 @@
+#!/usr/bin/env python3
+import sys
+
+alpha = {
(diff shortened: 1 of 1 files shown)


chall01/jsalmi.py
```diff
@@ -0,0 +1,37 @@
...
+        's' : "...",
+        't' : "-",
+        'u' : "..-",
+        'v' : "...-",
+        'w' : ".--",
+        'x' : "-..-",
+        'y' : "-.--",
+        'z' : "--..",
+        ' ' : " "}
+if len(sys.argv) == 2 and len(sys.argv[1]) > 0 and all(i.lower() in alpha for i in sys.argv[1]):
```
jgengo, line 32: I don't understand your last check condition, why?

chall01/jsalmi.py
```diff
@@ -0,0 +1,37 @@
...
+    for letter in sys.argv[1]:
+        print(alpha[letter.lower()], end = '')
+    print('')
```
jgengo, line 35: you should avoid printing one by one and append it to a string then print the string.

chall01/jsalmi.py
```diff
@@ -0,0 +1,37 @@
...
+else:
+    print ("usage:", sys.argv[0], "<a-zA-Z string>")
```
jgengo, line 37: good!

chall01/jsalmi.py
J0NY97: It is to check that all of the characters in the input string is  one of the characters in the "alpha" dictionary.

chall01/jsalmi.py
J0NY97: ok =)

chall01/jsalmi.py
J0NY97: Thanks :)

PR #72, opened 2020-03-26
chall01

DoebelnAtz:
chall01 solution aadlercr

### DIFF
--- a/chall01/aadlercr.py
+++ b/chall01/aadlercr.py
@@ -1,0 +1,57 @@
+#!/usr/bin/python3
+
+import sys
+
+
+def printMorse(message):
+    morse = [
+        '.-',
+        '-...',
+        '-.-.',
+        '-..',
+        '.',
+        '..-.',
+        '--.',
+        '....',
+        '..',
+        '.---',
+        '-.-',
+        '.-..',
+        '--',
+        '-.',
+        '---',
+        '.--.',
+        '--.-',
+        '.-.',
+        '...',
+        '-',
+        '..-',
+        '...-',
+        '.--',
+        '-..-',
+        '-.--',
+        '--..'
+    ]
+    res = ''
+    for letter in message:
+        if 97 <= ord(letter.lower()) <= 122:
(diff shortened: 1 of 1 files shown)


chall01/aadlercr.py
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/python3
+
+import sys
+
+
+def printMorse(message):
+    morse = [
```
jgengo, line 7: I would recommend put that out of the function, to avoid redeclaring/assigning this variable everytime you call printMorse()

chall01/aadlercr.py
```diff
@@ -0,0 +1,57 @@
...
+        '..-',
+        '...-',
+        '.--',
+        '-..-',
+        '-.--',
+        '--..'
+    ]
+    res = ''
+    for letter in message:
+        if 97 <= ord(letter.lower()) <= 122:
```
jgengo, line 37: you should use `letter.isalpha()`

chall01/aadlercr.py
```diff
@@ -0,0 +1,57 @@
...
+            res = res + morse[ord(letter.lower()) - 97]
+        else:
+            if ord(letter) == 32:
```
jgengo, line 40: to be clearer : if letter == " ":

chall01/aadlercr.py
```diff
@@ -0,0 +1,57 @@
...
+                res = res + ' '
+            else:
+                print("usage: ./aadlercr.py <a-zA-Z string>")
```
jgengo, line 43: It would be better to use sys.argv[0] than hardcode your login

chall01/aadlercr.py
```diff
@@ -0,0 +1,57 @@
...
+
+def main():
+    if not len(sys.argv) == 2:
+        print("usage: ./aadlercr.py <a-zA-Z string>")
+    else:
+        printMorse(sys.argv[1])
+    return
+
+
+if __name__ == '__main__':
```
jgengo, line 56: good!

PR #73, opened 2020-03-26
Chall01 by mhasan

mhjony:
1. Checked argc == 2
2. Declared a variable(type dictionary) that holds all the morse codes. 
3. Using for loop converted each character of the input string into morse code. 
4. Set a if condition for alphabet and space.
5. Retrun the encoded output.
6. Error handling

### DIFF
--- a/chall01/mhasan.py
+++ b/chall01/mhasan.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/python3
 
 import sys
 
--- a/chall01/mhasan.py
+++ b/chall01/mhasan.py
@@ -3,9 +3,8 @@
 import sys
 
 if len(sys.argv) != 2:
-    print("usage: ./xlogin.py <a-zA-Z string>")
+    print("usage: ./mhasan.py <a-zA-Z string>")
     sys.exit(1)
-
 
 def to_morse_code(input):
     code = {'a': '.-', 'b': '-...', 'c': '-.-.', 'd': '-..', 'e': '.', 'f': '..-.', 'g': '--.', 'h': '....',
@@ -20,12 +19,12 @@ def to_morse_code(input):
             else:
                 morse_code += code[x.lower()]
         else:
-            print("usage: ./xlogin.py <a-zA-Z string>")
+            print("usage: ./mhasan.py <a-zA-Z string>")
             sys.exit(1)
     return morse_code
 
 input = sys.argv[1]
 if (input == ""):
-    print("usage: ./xlogin.py <a-zA-Z string>")
+    print("usage: ./mhasan.py <a-zA-Z string>")
     sys.exit(1)
 print(to_morse_code(input))
--- a/chall01/mhasan.py
+++ b/chall01/mhasan.py
@@ -1,0 +1,31 @@
+#!/usr/bin/python
+
+import sys
+
+if len(sys.argv) != 2:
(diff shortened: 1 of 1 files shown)


chall01/mhasan.py
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/python3
+
+import sys
+
+if len(sys.argv) != 2:
+    print("usage: ./mhasan.py <a-zA-Z string>")
+    sys.exit(1)
```
d15ky, line 7: you can use sys.argv[0] for program name
also, "exit" takes a string as an argument, check what for :)

chall01/mhasan.py
```diff
@@ -0,0 +1,30 @@
...
+
+def to_morse_code(input):
```
d15ky, line 9: Huuuuge shoutout for making a separate function here. Makes total sense, well done!

chall01/mhasan.py
```diff
@@ -0,0 +1,30 @@
...
+    code = {'a': '.-', 'b': '-...', 'c': '-.-.', 'd': '-..', 'e': '.', 'f': '..-.', 'g': '--.', 'h': '....',
+            'i': '..', 'j': '.---', 'k': '-.-', 'l': '.-..', 'm': '--', 'n': '-.', 'o': '---', 'p': '.--.',
+            'q': '--.-', 'r': '.-.', 's': '...', 't': '-', 'u': '..-', 'v': '...-', 'w': '.--', 'x': '-..-',
+            'y': '-.--', 'z': '--..'}
```
d15ky, line 13: Using dictionary - smart! 
It also can be formed automatically by using split method of string
`morse_splitted = ".- -...".split(" ")`
and using zip to form a dict
`code = dict(zip(string.ascii_lowercase, morse_splitted))`

chall01/mhasan.py
```diff
@@ -0,0 +1,30 @@
...
+    morse_code = ""
+    for x in input:
+        if (x.isalpha() or x == ' '):
+            if (x == ' '):
+                morse_code += ' '
```
d15ky, line 18: would be more convenient to have " " in dictionary keys

chall01/mhasan.py
```diff
@@ -0,0 +1,30 @@
...
+            else:
+                morse_code += code[x.lower()]
+        else:
+            print("usage: ./mhasan.py <a-zA-Z string>")
+            sys.exit(1)
+    return morse_code
```
d15ky, line 24: There is a "pythonic" way to write Python code. Check it on the internet but basically it means to use all the possibilities of the language to write laconic code (still keeping in mind that it should be readable, etc.) 
So here is a pythonic way to write this loop in one line: 
`morse_code = "".join(code[x] for x in input.lower())`
Given that you'd checked that string doesn't have wrong chars: 
`if not input.lower().replace(" ", "").isalpha()`
and that your dict has " " key

chall01/mhasan.py
```diff
@@ -0,0 +1,30 @@
...
+
+input = sys.argv[1]
+if (input == ""):
+    print("usage: ./mhasan.py <a-zA-Z string>")
+    sys.exit(1)
+print(to_morse_code(input))
```
d15ky, line 30: Just want to admit once again how great your program structure is (I mean separate function to translate a string to morse code)

d15ky (approved):
Overall, clean and readable code. Nice solution, well done!

PR #74, opened 2020-03-26
chall01

ihwang:
Summary
1. Check the given argument has suitable form for this program.
2. Convert to morse sign(Print its usage, otherwise)

What I've learnt
1. Basic syntax of python
2. How to take command line argument into python script(sys module)
3. Functions 'range()', 'len()' and combinations with 'for'

### DIFF
--- a/chall01/ihwang.py
+++ b/chall01/ihwang.py
@@ -1,0 +1,76 @@
+#!/usr/bin/env python3
+
+import sys
+
+def intp_to_mos(av):
+	for i in range(len(av)):
+		if (av[i] == 'A') | (av[i] == 'a'):
+			print('.-', end='')
+		elif (av[i] == 'B') | (av[i] == 'b'):
+			print('-...', end='')
+		elif (av[i] == 'C') | (av[i] == 'c'):
+			print('-.-.', end='')
+		elif (av[i] == 'D') | (av[i] == 'd'):
+			print('-..', end='')
+		elif (av[i] == 'E') | (av[i] == 'e'):
+			print('.', end='')
+		elif (av[i] == 'F') | (av[i] == 'f'):
+			print('..-.', end='')
+		elif (av[i] == 'G') | (av[i] == 'g'):
+			print('--.', end='')
+		elif (av[i] == 'H') | (av[i] == 'h'):
+			print('....', end='')
+		elif (av[i] == 'I') | (av[i] == 'i'):
+			print('..', end='')
+		elif (av[i] == 'J') | (av[i] == 'j'):
+			print('.---', end='')
+		elif (av[i] == 'K') | (av[i] == 'k'):
+			print('-.-', end='')
+		elif (av[i] == 'L') | (av[i] == 'l'):
+			print('.-..', end='')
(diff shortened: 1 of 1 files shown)


chall01/ihwang.py
```diff
@@ -0,0 +1,76 @@
+#!/usr/bin/env python3
+
+import sys
+
+def intp_to_mos(av):
```
d15ky, line 5: Good job using a separate function

chall01/ihwang.py
```diff
@@ -0,0 +1,76 @@
...
+			print('.--', end='')
+		elif (av[i] == 'X') | (av[i] == 'x'):
+			print('-..-', end='')
+		elif (av[i] == 'Y') | (av[i] == 'y'):
+			print('-.--', end='')
+		elif (av[i] == 'Z') | (av[i] == 'z'):
+			print('--..', end='')
+		elif av[i] == ' ':
+		 	print(av[i], end='')
+	print('')
```
d15ky, line 61: Instead of using a lot of if-else statements, it's better to use dictionaries. It's more "pythonic" (google it:) way to do things.

chall01/ihwang.py
```diff
@@ -0,0 +1,76 @@
...
+		
+av = sys.argv[1:]
+if len(av) == 1:
+	if len(av[0]) == 0:
+		print('usage: ', sys.argv[0], '<a-zA-Z string>')
```
d15ky, line 66: Usage of sys.argv[0] - great!

chall01/ihwang.py
```diff
@@ -0,0 +1,76 @@
+#!/usr/bin/env python3
+
+import sys
+
+def intp_to_mos(av):
+	for i in range(len(av)):
```
d15ky, line 6: no need to iterate by index in Python in this case
you can do it like that:
`for c in av:`
and c will be each character of av (or each element if it's a list or dict)

chall01/ihwang.py
```diff
@@ -0,0 +1,76 @@
...
+	print('')
+		
+av = sys.argv[1:]
+if len(av) == 1:
+	if len(av[0]) == 0:
+		print('usage: ', sys.argv[0], '<a-zA-Z string>')
+		exit()
+	for i in range(len(av[0])):
+		if not(('A' <= av[0][i] <= 'Z') | ('a' <= av[0][i] <= 'z')  \
+				| (av[0][i] == ' ')):
```
d15ky, line 70: There is a method in python string object that allows checking if it's an alpha kind: 
`if not (av[0][i].isalpha() or av[0][i] == ' '):`

d15ky (approved):
Well done discovering Python! Great job for the first day of Python. Below are suggestions where to move next:
Your Python code is very much like a C code and you don't use the possibilities Python provides. It's general advice about learning new languages or working when you are allowed to use all the functions. Try to understand what is the philosophy of the language and try to ease your manual work as much as possible (without stopping paying attention to efficiency, of course). I think it's what 42 program students struggle the most with - using shortcuts when they are allowed to. 

But anyway, that's something to consider when you'll be preparing for the first internship. For now - good job :)

PR #75, opened 2020-03-26
asolopov asolution

Artemso:
I've created a dict for the morse code, where each lowercase letter corresponds to the code.

1) I check for the number of arguments and if needed argument is empty or not.
2) I check for allowed chars: alphabetic and spaces, I also check if the letter is present in the dict.
3) Put the string into lowercase.
4) For each character in the string:
    if its a letter-> print corresponding value for its key.
    if its a space-> print space.
5) Print newline.

In case of error( wrong arg num or bad chars) I print usage.


Artemso:
Two bugs I've realised:
1) I should manage alphabetic letters like cyrillic(since they are alphabetical and not present in the morse dictionary)
2) I didn't notice usage has program name written ^_^

Artemso:
Yep, check for the empty string. 

¯\\_(ツ)_/¯

Artemso:
I've noticed in slack that the file has to be executable, didn't know that before so I learned some new stuff again! :)

chall01/asolopov.py
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/python3
+import sys
+
+morse_dict = {
```
plantran, line 4: Very nice implementation of the dict with the letters as keys. 👍

chall01/asolopov.py
```diff
@@ -0,0 +1,58 @@
...
+			if x not in morse_dict:
+				usage()
+
+if (len(sys.argv)) != 2 or not sys.argv[1]:
+	usage()
+else:
+	word = sys.argv[1].lower()
+	check_chars(word, morse_dict)
+	for x in word:
+		if (x.isalpha()):
```
plantran, line 52: One suggestion is to add the space key/value pair in your morse_dict. That would allow you to remove that check of isalpha or isspace. Same as in your `check_chars()` function, it would prevent you this different checks, you could then just do 
```
if x not in morse_dict:
  usage()
```

chall01/asolopov.py
```diff
@@ -0,0 +1,58 @@
...
+"u": "..-",
+"v": "...-",
+"w": ".--",
+"x": "-..-",
+"y": "-.--",
+"z": "--.."
+}
+
+def usage():
+	print(f"usage: {sys.argv[0]} <a-zA-Z string>")
```
plantran, line 34: Nice usage of `sys.argv[0]` for the program name.

plantran:
Hello!

I've added some comments and suggestions directly in your file. I will just add here that it would have been great to protect your script from running on import by adding `if __name__ == "__main__":`. You can check more about it [here](https://stackoverflow.com/questions/419163/what-does-if-name-main-do/419185#419185).

Otherwise, the solution is very clear and the pull request description understandable, I'm merging the pull request, good job 👍

PR #77, opened 2020-03-26
rjaakonm string to morse

rasmusjaa:
1. Create list with all morse letter codes.
2. Start main function that checks there's only 2 argument and [1] is not empty. 
3. Create empty string and change arg[1] to lowercase. 
4. Loop through arg[1] and get ascii minus 97 from all alphabets (that are now lowercase) to find element from list and concat that to string. If space, concat space, if anything else, show usage. 
5. Print string.

### DIFF
--- a/chall01/rjaakonm.py
+++ b/chall01/rjaakonm.py
@@ -30,10 +30,10 @@ morse = [".-",
 	"--.."]
 
 def main():
-	str = ""
 	if (len(sys.argv) != 2) or (len(sys.argv[1]) == 0):
 		print("usage: ./rjaakonm.py <a-zA-Z string>")
 		return
+	str = ""
 	sys.argv[1] = sys.argv[1].lower()
 	for char in sys.argv[1]:
 		if char.isalpha():
--- a/chall01/rjaakonm.py
+++ b/chall01/rjaakonm.py
@@ -1,0 +1,49 @@
+#!/usr/bin/python3
+
+import sys
+
+morse = [".-",
+	"-...",
+	"-.-.",
+	"-..",
+	".",
+	"..-.",
+	"--.",
+	"....",
+	"..",
+	".---",
+	"-.-",
+	".-..",
+	"--",
+	"-.",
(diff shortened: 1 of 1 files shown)


chall01/rjaakonm.py
```diff
@@ -0,0 +1,49 @@
...
+		if char.isalpha():
+			str += morse[ord(char) - 97]
+		elif char == " ":
+			str += " "
+		else:
+			print("usage: ./rjaakonm.py <a-zA-Z string>")
+			return
+	print(str)
+
+if __name__ == "__main__":
```
plantran, line 48: Very good practice with the `if __name__ == "__main__"` 👍

chall01/rjaakonm.py
```diff
@@ -0,0 +1,49 @@
...
+	"..-",
+	"...-",
+	".--",
+	"-..-",
+	"-.--",
+	"--.."]
+
+def main():
+	if (len(sys.argv) != 2) or (len(sys.argv[1]) == 0):
+		print("usage: ./rjaakonm.py <a-zA-Z string>")
```
plantran, line 34: You should prefer using `sys.argv[0]` to display your program name.

chall01/rjaakonm.py
```diff
@@ -0,0 +1,49 @@
...
+		return
+	str = ""
+	sys.argv[1] = sys.argv[1].lower()
+	for char in sys.argv[1]:
+		if char.isalpha():
+			str += morse[ord(char) - 97]
```
plantran, line 40: This solution works but if I can give you an advice, you should check python `dict`. If you're familiar with the concept of a "hash table" it's basically the same. It woud have allowed you to associate the letters with the corresponding morse code. For example `morse['a'] -> ".-"`.

plantran:
Hi!

I've added some comment and suggestions to your file.
Otherwise, your code is clean and the PR description is understandable. Good usage of the shebang and the main. Good job, I'm merging this pull request 👍

PR #78, opened 2020-03-26
tkuumola solution chall01

T7Q:
1. Created a dictionary with morse code + empty space.
2. Checked that there are 2 arguments, if not error message is shown
3. Created variable "input". Saved capitalized content of argv[1] to it.
4. Created variable "len" to measure length of string and identify empty or full of spaces string with split and join functions.
4. If statement to check if "input" string is not empty and contains only letters based on morse_code keys.
- If condition is met loop though the elements of the "input" string and print values from the dictionary.
- Error message in other cases.

### DIFF
--- a/chall01/tkuumola.py
+++ b/chall01/tkuumola.py
@@ -1,0 +1,49 @@
+#!/usr/bin/env python3
+
+import sys
+
+morse_code = {
+    'A':'.-',
+    'B':'-...',
+    'C':'-.-.',
+    'D':'-..',
+    'E':'.',
+    'F':'..-.',
+    'G':'--.',
+    'H':'....',
+    'I':'..',
+    'J':'.---',
+    'K':'-.-',
+    'L':'.-..',
+    'M':'--',
+    'N':'-.',
+    'O':'---',
+    'P':'.--.',
+    'Q':'--.-',
+    'R':'.-.',
+    'S':'...',
+    'T':'-',
+    'U':'..-',
+    'V':'...-',
+    'W':'.--',
+    'X':'-..-',
+    'Y':'-.--',
+    'Z':'--..',
+    ' ':' ',
+}
+
+if len(sys.argv) != 2:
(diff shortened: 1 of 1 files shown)


chall01/tkuumola.py
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env python3
```
plantran, line 1: 👍

chall01/tkuumola.py
```diff
@@ -0,0 +1,47 @@
...
+
+import sys
+
+morse_code = {
```
plantran, line 5: I really appreciate your usage of a dict and your implementation with the letters as keys is well thought!

chall01/tkuumola.py
```diff
@@ -0,0 +1,47 @@
...
+    'V':'...-',
+    'W':'.--',
+    'X':'-..-',
+    'Y':'-.--',
+    'Z':'--..',
+    ' ':' ',
+}
+
+if len(sys.argv) != 2:
+    print ('usage: ./tkuumola.py <a-zA-Z string>')
```
plantran, line 36: You should prefer using `sys.argv[0]` to display your program name.

chall01/tkuumola.py
```diff
@@ -0,0 +1,47 @@
...
+    sys.exit(1)
+
+string = sys.argv[1].upper()
+str_len = len(" ".join(string.split()))
+
+if all(x in morse_code for x in string) and str_len != 0:
```
plantran, line 42: Good check

plantran:
Hi!

I've added specific comments and suggestions to your code. I will add here that you should protect your script with `if __name__ == '__main__':` which will serve as an entry-point. You can read more about this [here](https://stackoverflow.com/questions/419163/what-does-if-name-main-do/419185#419185).

Other than this, this is a good job, the code is clean and the pull-request description is understandable. I'm merging this pull request 👍

PR #79, opened 2020-03-26
My implementation of 'String to Morse'

DianaMukaliyeva:
1. Checked that number of arguments equal 2 and first argument is not empty.
2. Created Python dictionary with keys as letters and values as Morse code.
3. Checked whether each character in the given string is in dictionary. If not  - printed usage message and exited from program, otherwise printed encoded string.


chall01/dmukaliy.py
```diff
@@ -0,0 +1,26 @@
...
+def morse(string):
+	alphabet = {" ":" ", "a":".-", "b":"-...", "c":"-.-.", "d":"-..", "e":".", "f":"..-.", "g":"--.", "h":"....",
+		"i":"..", "j":".---", "k":"-.-", "l":".-..", "m":"--", "n":"-.", "o":"---", "p":".--.", "q":"--.-",
+		"r":".-.", "s":"...", "t":"-", "u":"..-", "v":"...-", "w":".--", "x":"-..-", "y":"-.--", "z":"--.."}
+	for char in string:
+		if char not in alphabet.keys():
+			error()
+	for char in string:
+		sys.stdout.write(alphabet[char])
+	sys.stdout.write("\n")
```
jgengo, line 17: you should avoid printing char by char but append alphabet[char] to a string and then print the string

chall01/dmukaliy.py
```diff
@@ -0,0 +1,26 @@
...
+
+def main():
+	if len(sys.argv) != 2 or not sys.argv[1]:
+		error() 
+	else:
+		morse(sys.argv[1].lower())
+
+if __name__ == "__main__":
+	main()
```
jgengo, line 26: Really good you protected it!

chall01/dmukaliy.py
```diff
@@ -0,0 +1,26 @@
...
+	sys.stdout.write("usage: ./dmukaliy.py <a-zA-Z string>\n")
+	exit()
+
+def morse(string):
+	alphabet = {" ":" ", "a":".-", "b":"-...", "c":"-.-.", "d":"-..", "e":".", "f":"..-.", "g":"--.", "h":"....",
+		"i":"..", "j":".---", "k":"-.-", "l":".-..", "m":"--", "n":"-.", "o":"---", "p":".--.", "q":"--.-",
+		"r":".-.", "s":"...", "t":"-", "u":"..-", "v":"...-", "w":".--", "x":"-..-", "y":"-.--", "z":"--.."}
+	for char in string:
+		if char not in alphabet.keys():
+			error()
```
jgengo, line 14: It's good! and if you add value to your alphabet dict it will keep working.

You could have been faster with 
```python
if not string.replace(" ", "").isalpha():
  error()
```

chall01/dmukaliy.py
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/python3
+import sys
+
+def error():
+	sys.stdout.write("usage: ./dmukaliy.py <a-zA-Z string>\n")
```
jgengo, line 5: you should avoid hardcoding your script name, you can use `sys.argv[0]` to find the script name

chall01/dmukaliy.py
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/python3
+import sys
+
+def error():
```
jgengo, line 4: it's really DRY concept to create a function error as you know you will use it many times. I would just maybe recommend using another name than: error() which is too generic by something like `print_usage()` or `print_and_exit(string)`

PR #80, opened 2020-03-26
My simple approach

RandommUser:
Morse code is made into an associative array. The code parses thru all the letters in the string, tests they are alphabets, transforms them into the index and gets the morse code representation of the letter. This is appended to the return string. Spaces are handled in their own case. Any invalid input will make the program print the usage instead. Once the string has been parsed it gets printed out and the program exits.

The program only accepts one parameter and does not accept and empty string.

### DIFF
--- a/chall01/phakakos.py
+++ b/chall01/phakakos.py
@@ -1,6 +1,8 @@
 import sys
 import string
 
+def pUsage():
+        print("usage: ./phakakos.py <a-zA-Z string>")
 
 def morse_letter(num):
 	num+=1
@@ -42,11 +44,15 @@ def morse_word(word):
                 elif (c == ' '):
                         morseWord+=' '
                 else:
-                        print("usage: ./phakakos.py <a-zA-Z string>")
+                        pUsage()
                         return
-        print(morseWord)
+        if (morseWord == ""):
+                pUsage()
+                return
+        else:
+                print(morseWord)
 
 if len(sys.argv) == 2:
 	morse_word(sys.argv[1])
 else:
-	print("usage: ./phakakos.py <a-zA-Z string>")
+	pUsage()
--- a/chall01/phakakos.py
+++ b/chall01/phakakos.py
@@ -40,7 +40,11 @@ def morse_word(word):
         morseWord =""
         for c in word:
                 if (c.isalpha()) == True:
-                        morseWord+=morse_letter(string.ascii_lowercase.index(c.lower()))
+                        morseRet =morse_letter(string.ascii_lowercase.index(c.lower()))
+                        if (morseRet == "invalid"):
+                                pUsage()
+                                return
+                        morseWord+=morseRet
                 elif (c == ' '):
                         morseWord+=' '
                 else:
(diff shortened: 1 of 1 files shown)


RandommUser:
Added a regex matching to prevent any unwanted letters to pass to functions that do not support them and to not parse strings that do not match the wanted input

chall01/phakakos.py
```diff
@@ -0,0 +1,61 @@
...
+
+pattern = re.compile("^[a-zA-Z ]+$")
+
+def pUsage():
+        print("usage: ./phakakos.py <a-zA-Z string>")
+        exit(1)
+
+def morse_letter(num):
+	num+=1
+	code={
```
plantran, line 13: Instead of having numbers as keys in your dict, you could have added the letters directly. For example:
```
code = {'a': '.-', 'b': '-...'', etc}
```.

File: chall01/phakakos.py [hunk 2]
```diff
@@ -0,0 +1,61 @@
...
+	}
+	return code.get(num, "invalid")
+
+def morse_word(word):
+        if (pattern.match(word) == None):
+                pUsage()
+        morseWord =""
+        for c in word:
+                if (c.isalpha()) == True:
+                        morseWord+=morse_letter(string.ascii_lowercase.index(c.lower()))
```
plantran, line 49: By changing your dict keys with the letter directly, you could have prevented `string.ascii_lowercase.index(`.
Calling `code['a']` will provide you with the right value.

chall01/phakakos.py
```diff
@@ -0,0 +1,61 @@
+import sys
+import string
+import re
+
+pattern = re.compile("^[a-zA-Z ]+$")
+
+def pUsage():
+        print("usage: ./phakakos.py <a-zA-Z string>")
```
plantran, line 8: You should prefer using `sys.argv[0]` to display the program name, it's a good habit.

chall01/phakakos.py
```diff
@@ -0,0 +1,61 @@
...
+                else:
+                        pUsage()
+        print(morseWord)
+
+if len(sys.argv) == 2:
+	morse_word(sys.argv[1])
+else:
+	pUsage()
+
+exit()
```
plantran, line 61: You don't need to add this `exit()` at the end. The python script will end once it will have no other instruction.

plantran:
Hi!

I've added some suggestions in your code. I will also add this:

- You should add a shebang at the top of your file so it can be executed directly from the command-line without calling the `python` binary first. (`#!/usr/bin/env python3`)
- You should protect your script with if __name__ == '__main__': which will serve as an entrypoint. You can read more about this [here](https://stackoverflow.com/questions/419163/what-does-if-name-main-do/419185#419185).

Other than that, good job! I like that you thought about using a dict for this assigment, that was a good thought. I'm merging this pull request 👍

chall01/phakakos.py
RandommUser: Never thought about that, so now I know to use it in my bash scripts too.

PR #81, opened 2020-03-26
Added vgrankul.py

VictoriaElisabet:
My first function made with Python. :)
I created an array for the morse codes and the alphabets. 
I will start by checking if the string contains other things than the alphabet, I will print usage if the string contains incorrect info.
Otherwise I will go into a while loop and increase i until i get the same letter as the letter from the string, based on that I will choose the morsecode from the array and add it to my sos_string. 
The sos_string will after the argvstring has been looped through be printed.

### DIFF
--- a/chall01/vgrankul.py
+++ b/chall01/vgrankul.py
@@ -1,0 +1,50 @@
+#!/bin/python3
+import sys
+
+sos = [
+".-",
+"-...", 
+"-.-.", 
+"-..", 
+".", 
+"..-.", 
+"--.", 
+"....", 
+"..", 
+".---", 
+"-.-", 
+".-..", 
+"--", 
+"-.", 
+"---", 
+".--.", 
+"--.-", 
+".-.", 
+"...", 
+"-", 
+"..-", 
+"...-", 
+".--", 
+"-..-", 
+"-.--", 
+"--.."]
+
+i = 0
+letter = ['a', 'b','c','d','e','f','g','h','i','j','k','l','m','n','o','p','q','r','s','t','u','v','w','x','y','z']
+Bletter = ['A', 'B','C','D','E','F','G','H','I','J','K','L','M','N','O','P','Q','R','S','T','U','V','W','X','Y','Z']
+sos_string = ""
+
+if len(sys.argv) != 2:
+    print("usage: ./vgrankul.py <a-zA-Z string>")
+else:
+    string = sys.argv[1]
+    if string.replace(' ', '').isalpha() == False:
+        print("usage: ./vgrankul.py <a-zA-Z string>")
+        exit()
+    for let in string:
+            i = 0
+            while i < 26 and i < 26:
+                if letter[i] == let or Bletter[i] == let:
+                    sos_string += sos[i]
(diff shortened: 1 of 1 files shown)


chall01/vgrankul.py
```diff
@@ -0,0 +1,50 @@
...
+"--.-", 
+".-.", 
+"...", 
+"-", 
+"..-", 
+"...-", 
+".--", 
+"-..-", 
+"-.--", 
+"--.."]
```
jgengo, line 30: I suggest declaring / assigning array like that one in a one-line like you did for letter or Bletter under.

chall01/vgrankul.py
```diff
@@ -0,0 +1,50 @@
...
+Bletter = ['A', 'B','C','D','E','F','G','H','I','J','K','L','M','N','O','P','Q','R','S','T','U','V','W','X','Y','Z']
+sos_string = ""
+
+if len(sys.argv) != 2:
+    print("usage: ./vgrankul.py <a-zA-Z string>")
+else:
+    string = sys.argv[1]
+    if string.replace(' ', '').isalpha() == False:
+        print("usage: ./vgrankul.py <a-zA-Z string>")
+        exit()
```
jgengo, line 43: It's the perfect way to solve this.

However, you could have written:

```python
if not string.replace(' ', '').isalpha()
```
Plus, the convention about printing `usage` is:
- if you have to write the usage because of an input error/user error you print the usage to stderr
- if the user as request help with for example : /yourScript.py --help then you print it to stdout.

chall01/vgrankul.py
```diff
@@ -0,0 +1,50 @@
...
+
+i = 0
+letter = ['a', 'b','c','d','e','f','g','h','i','j','k','l','m','n','o','p','q','r','s','t','u','v','w','x','y','z']
+Bletter = ['A', 'B','C','D','E','F','G','H','I','J','K','L','M','N','O','P','Q','R','S','T','U','V','W','X','Y','Z']
+sos_string = ""
+
+if len(sys.argv) != 2:
+    print("usage: ./vgrankul.py <a-zA-Z string>")
+else:
+    string = sys.argv[1]
```
jgengo, line 40: you could have used `.lower()` to sys.argv[1] before processing to avoid having to handle both and having 2 arrays

chall01/vgrankul.py
```diff
@@ -0,0 +1,50 @@
...
+"-", 
+"..-", 
+"...-", 
+".--", 
+"-..-", 
+"-.--", 
+"--.."]
+
+i = 0
+letter = ['a', 'b','c','d','e','f','g','h','i','j','k','l','m','n','o','p','q','r','s','t','u','v','w','x','y','z']
```
jgengo, line 33: or 
```python
list(string.ascii_lowercase)
```

jgengo (commented):
For your first time with Python you can be proud of you.

It's done by
```python
if __name__ == "__main__":
```

VictoriaElisabet:
Thank you very much for taking your time to give all that feedback ! :)

PR #82, opened 2020-03-27
StringtoMorse

tangxing1220:
This is my first PR and my  first python code.

### DIFF
--- a/chall01/xtang.py
+++ b/chall01/xtang.py
@@ -1,0 +1,59 @@
+#!/usr/bin/python3
+
+import sys
+
+def errorMessage():
+	print(f"usage: {sys.argv[0]} <a-zA-Z string>")
+	exit()
+
+morse_dict = {
+"a": ".-",
+"b": "-...",
+"c": "-.-.",
+"d": "-..",
+"e": ".",
+"f": "..-.",
+"g": "--.",
+"h": "....",
+"i": "..",
+"j": ".---",
+"k": "-.-",
+"l": ".-..",
+"m": "--",
+"n": "-.",
+"o": "---",
+"p": ".--.",
+"q": "--.-",
+"r": ".-.",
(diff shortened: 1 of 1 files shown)


chall01/xtang.py
```diff
@@ -0,0 +1,59 @@
...
+"r": ".-.",
+"s": "...",
+"t": "-",
+"u": "..-",
+"v": "...-",
+"w": ".--",
+"x": "-..-",
+"y": "-.--",
+"z": "--.."
+}
```
jgengo, line 36: dict was a good idea!

chall01/xtang.py
```diff
@@ -0,0 +1,59 @@
...
+
+def check_chars(word, morse_dict):
+	for x in word:
+		if x.isalpha() == False:
+			if x.isspace() == False:
+				errorMessage()
+		else:
+			if x not in morse_dict:
+				errorMessage()
```
jgengo, line 45: isalpha() can be used in a string or a char 

you also know that your dict keys is all the a-z lowercase.

I would recommend changing all of this by

```python
if not word.replace(" ", "").isalpha():
  errorMessage()
```

chall01/xtang.py
```diff
@@ -0,0 +1,59 @@
...
+if (len(sys.argv)) != 2 or not sys.argv[1]:
+	errorMessage()
+else:
+	word = sys.argv[1].lower()
+	check_chars(word, morse_dict)
+	for x in word:
+		if (x.isalpha()):
+			print(morse_dict[x], end='')
+		elif (x.isspace()):
+			print(" ", end='')
```
jgengo, line 56: Here, I would recommend avoiding printing char by char but append to a string every char and then print it.

There are more pythonic ways to write this also.

jgengo (approved):

PR #83, opened 2020-03-27
first time forking and submitting pull requests

rikikoo:


chall01/rkyttala.py
```diff
@@ -0,0 +1,69 @@
...
+If valid, convert input to lowercase.
+"""
+if len(sys.argv) == 2 and sys.argv[1].replace(" ", "").isalpha():
+    if not re.match(r"^[a-z A-Z]+$", sys.argv[1]):
+        print("usage: ./rkyttala.py <a-zA-Z string>")
+        quit()
+    else:
+        progInput = sys.argv[1].lower()
+else:
+    print("usage: ./rkyttala.py <a-zA-Z string>")
```
plantran: You should rather use `sys.argv[0]` to display the name of your program.

chall01/rkyttala.py
```diff
@@ -0,0 +1,69 @@
...
+
+"""
+Check that 
+- there's only one argument,
+- argument contains only a to z (no ä, ö, å, etc.) and spaces.
+If valid, convert input to lowercase.
+"""
+if len(sys.argv) == 2 and sys.argv[1].replace(" ", "").isalpha():
+    if not re.match(r"^[a-z A-Z]+$", sys.argv[1]):
+        print("usage: ./rkyttala.py <a-zA-Z string>")
```
plantran, line 60: You should rather use `sys.argv[0]` to display your program name.

chall01/rkyttala.py
```diff
@@ -0,0 +1,69 @@
...
+                - 
+                ..- 
+                ...- 
+                .-- 
+                -..- 
+                -.-- 
+                --.."""
+
+    rawAlphabet = "a b c d e f g h i j k l m n o p q r s t u v w x y z"
+    alphabet = rawAlphabet.split()
```
plantran, line 36: If you're going to split both strings anyway, you should declare them both as arrays directly!

chall01/rkyttala.py
```diff
@@ -0,0 +1,69 @@
...
+    morse = rawMorse.replace(" ", "").split("\n")
+
+    # merge alphabet and morse lists into a dictionary
+    codeBook = dict(zip(alphabet, morse))
```
plantran, line 40: You could have already initialized your `codeBook` as a dict.

plantran:
Hello!

Unfortunately, you didn't provide any description with your pull request explaining your code and your way of resolving the assignment so I won't merge it.

I've added some suggestions to your code, check above. 
You should protect your script with `if __name__ == '__main__':`  which will serve as an entrypoint. You can read more about this [here](https://stackoverflow.com/a/419185/12108768).

Other than that good job 👍

PR #84, opened 2020-03-30
add my solution

salahadawi:
I malloc a pointer where the string created by sprintf will be stored. With sprintf, i use the format tags to convert each number to hexadecimal, with a minimum length of 2 digits. This string is then returned.

I took this approach as i recently created my own sprintf, and though that it would be a convenient way of solving this exercise.

### DIFF
--- a/chall02/sadawi.c
+++ b/chall02/sadawi.c
@@ -4,7 +4,7 @@
 
 char	*hv_rgb2hex(int r, int g, int b)
 {
-	char *str = malloc(1000);
+	char *str = malloc(8);
 	sprintf(str, "#%02x%02x%02x\n", (r), (g), b);
 	return (str);
 }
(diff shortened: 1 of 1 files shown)


chall02/sadawi.c
```diff
@@ -0,0 +1,10 @@
+
+#include <stdio.h>
+#include <stdlib.h>
+
+char	*hv_rgb2hex(int r, int g, int b)
+{
+	char *str = malloc(8);
```
jgengo, line 7: It would have been better if you would have protected your malloc.

chall02/sadawi.c
```diff
@@ -0,0 +1,10 @@
...
+	sprintf(str, "#%02x%02x%02x", r, g, b);
```
jgengo, line 8: Smart and good usage of sprintf ;)

jgengo (approved):
Really nice! 

That's a pity you did not protect your malloc, or try to check the input are at least between 0 and 255.

PR #85, opened 2020-03-30
Add my solution

hiiri:
**base16** - Contains hexadecimal digits 0-F
**color** = Red(0-255) * 65536 + Green(0-255) * 256 + Blue(0-255)
**idx** - Current hexadecimal digit (also index of string, starting from least significant digit which is _blue_)

By using **color** I can iterate over one variable. Calculate color mod 16 for the base 10 value for current hexadecimal digit, then divide by 16 and repeat until 6 hexadecimal digits have been calculated. Add a '#' to the start and return formed string.

Possible improvements: Bit shifting is probably much better for this task.


chall02/alcohen.c
```diff
@@ -0,0 +1,31 @@
...
+    }
+    s[0] = '#';
+    return (s);
+}
+
+char    *hv_rgb2hex(int r, int g, int b)
+{
+    int color;
+
+    if (r < 0 || r > 255 || g < 0 || g > 255 || b < 0 || b > 255)
```
jgengo, line 27: it's good you checked the input!

chall02/alcohen.c
```diff
@@ -0,0 +1,31 @@
...
+        return (NULL);
+    color = (r * 256 * 256) + (g * 256) + b;
```
jgengo, line 29: smart, where did you find this?

jgengo (approved):
Good job!

PR #86, opened 2020-03-30
RGB to hex

kimsappi:
I first check if all the values fall within 0-255. Then I start the allocated string with '#' and convert each value to hex with ft_itoa_base from my printf and concatenate that onto the original string. If the value is only one hex digit, I add a 0 before it.

### DIFF
--- a/chall02/ksappi.c
+++ b/chall02/ksappi.c
@@ -1,0 +1,82 @@
+#include <string.h>
+#include <stdlib.h>
+
+char	*ft_strnew(size_t size)
+{
+	char	*area;
+	size_t	i;
+
+	area = (char *)malloc(sizeof(char) * (size + 1));
+	if (!area)
+		return (NULL);
+	i = 0;
+	while (i < size + 1)
+	{
+		area[i] = '\0';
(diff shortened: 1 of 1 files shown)


chall02/ksappi.c
```diff
@@ -0,0 +1,82 @@
...
+{
+  if (color < 0 || color > 255)
+    return 0;
+  else
+    return 1;
+}
+
+char *hv_rgb2hex(int r, int g, int b)
+{
+  char *ret = (char *)malloc(8);
```
jgengo, line 56: this malloc isn't protected.

chall02/ksappi.c
```diff
@@ -0,0 +1,82 @@
...
+		--len;
+	}
+	return (str);
+}
+int check_if_valid_value(int color)
+{
+  if (color < 0 || color > 255)
+    return 0;
+  else
+    return 1;
```
jgengo, line 51: could have been one-lined, but I appreciate you checked the input given by the user.

jgengo (approved):
Good job!

PR #87, opened 2020-03-30
RGB to Hexadecimal

vossr:
Malloc result string, then for each color with sprintf convert int to hex.

### DIFF
--- a/chall02/rpehkone.c
+++ b/chall02/rpehkone.c
@@ -1,0 +1,22 @@
+#include <stdlib.h>
+#include <stdio.h>
+
+char	*hv_rgb2hex(int r, int g, int b)
+{
+	char *res;
+	char hex[5];
+
+	res = (char*)malloc(sizeof(char) * 8);
+	res[0] = '#';
+	sprintf(&hex[0], "%04x", r);
+	res[1] = hex[2];
+	res[2] = hex[3];
+	sprintf(&hex[0], "%04x", g);
+	res[3] = hex[2];
+	res[4] = hex[3];
+	sprintf(&hex[0], "%04x", b);
+	res[5] = hex[2];
+	res[6] = hex[3];
+	res[7] = '\0';
+	return (res);
+}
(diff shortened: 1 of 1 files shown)


chall02/rpehkone.c
```diff
@@ -0,0 +1,24 @@
+#include <stdlib.h>
+#include <stdio.h>
+
+char	*hv_rgb2hex(int r, int g, int b)
+{
+	char *res;
+	char hex[3];
+
+	if (r < 0 || g < 0 || b < 0 || r > 255 || g > 255 || b > 255)
```
jgengo, line 9: nice you checked the user input.

a small function `color_is_valid(int c)` would make it cleaner

chall02/rpehkone.c
```diff
@@ -0,0 +1,24 @@
...
+	sprintf(&hex[0], "%02x", r);
+	res[1] = hex[0];
+	res[2] = hex[1];
+	sprintf(&hex[0], "%02x", g);
+	res[3] = hex[0];
+	res[4] = hex[1];
+	sprintf(&hex[0], "%02x", b);
+	res[5] = hex[0];
+	res[6] = hex[1];
+	res[7] = '\0';
```
jgengo, line 22: nice you thought about using sprintf.

You could have just write `sprintf(hex, "#%02x%02x%02x, r, g, b);` I think.

chall02/rpehkone.c
```diff
@@ -0,0 +1,24 @@
...
+#include <stdio.h>
+
+char	*hv_rgb2hex(int r, int g, int b)
+{
+	char *res;
+	char hex[3];
+
+	if (r < 0 || g < 0 || b < 0 || r > 255 || g > 255 || b > 255)
+		return (NULL);
+	res = (char*)malloc(sizeof(char) * 8);
```
jgengo, line 11: Always protect your malloc

jgengo (approved):
nice, you used sprintf. Don't forget to always protect your malloc.

PR #88, opened 2020-03-30
rgb2hex

rasmusjaa:
1. Malloc string for hex
2. Clamp ints to 0-255
3. Use bitwise operations to set int values to hex
4. Sprintf # and hex to string
5. Return string

### DIFF
--- a/chall02/rjaakonm.c
+++ b/chall02/rjaakonm.c
@@ -1,0 +1,29 @@
+#include <stdio.h>
+#include <stdlib.h>
+
+int		ft_int_clamp(int nb, int min, int max)
+{
+	if (nb < min)
+		return (min);
+	if (nb > max)
+		return (max);
+	return (nb);
+}
+
+char *hv_rgb2hex(int r, int g, int b)
+{
+	int				rgb[3];
+	unsigned int 	color;
+	char			*color_string;
+
+	if (!(color_string = (char *)malloc(sizeof(char) * 7)))
+		return NULL;
+
+	rgb[0] = ft_int_clamp(r, 0, 255);
+	rgb[1] = ft_int_clamp(g, 0, 255);
+	rgb[2] = ft_int_clamp(b, 0, 255);
+	color = rgb[0] << 16 | rgb[1] << 8 | rgb[2];
+	sprintf(&color_string[1], "%x", color);
+	color_string[0] = '#';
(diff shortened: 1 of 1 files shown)


chall02/rjaakonm.c
```diff
@@ -0,0 +1,28 @@
...
+#include <stdlib.h>
+
+int		ft_int_clamp(int nb, int min, int max)
+{
+	if (nb < min)
+		return (min);
+	if (nb > max)
+		return (max);
+	return (nb);
+}
```
jgengo, line 11: That's really nice and DRY (don't repeat yourself)!

chall02/rjaakonm.c
```diff
@@ -0,0 +1,28 @@
...
+	int				rgb[3];
+	unsigned int 	color;
+	char			*color_string;
+
+	if (!(color_string = (char *)malloc(sizeof(char) * 8)))
+		return NULL;
+
+	rgb[0] = ft_int_clamp(r, 0, 255);
+	rgb[1] = ft_int_clamp(g, 0, 255);
+	rgb[2] = ft_int_clamp(b, 0, 255);
```
jgengo, line 24: that's really good, but I think your functions (all of them) should always do what they are meant for, or return an error if a user input isn't good. (or return NULL in this case).

chall02/rjaakonm.c
```diff
@@ -0,0 +1,28 @@
...
+	color = rgb[0] << 16 | rgb[1] << 8 | rgb[2];
```
jgengo, line 25: bit wise :) where did you find this idea?

chall02/rjaakonm.c
```diff
@@ -0,0 +1,28 @@
...
+	sprintf(color_string, "#%x", color);
```
jgengo, line 26: if you are using sprintf(); you could have avoided doing any bitwise by just doing

```c
sprintf(color_string, "#%02x%02x%02x", r, g, b);
```

jgengo (approved):

chall02/rjaakonm.c
rasmusjaa: I've been using it for all graphic branch stuff, was the fastest and shortest way I found to do those, but yeah wasn't useful for this since I still used sprintf :p but works great if I need to manipulate colors!

PR #89, opened 2020-03-30
Add krusthol's solution for RGB to Hexadecimal

krusthol:
Approach:
Since I have done similar things, I just combined some previous approaches I've used in the C Exams for hex conversion with the RGB value things I've coded in the graphics branch, added an error behaviour which was not asked for, but which I felt useful to implement anyway. Only one thing in code that bypasses Norm, the hex value map instanciation at declaration, and not including header by convention.

Code:
Simple hex conversion by modulo from Hex string table, formatting three color portions RGB, returning "#error!" string if out of range 0-255 for any of the RGB values. Mallocing the return string with malloc from stdlib.h, assuming memory management will be handled by the outer program.

### DIFF
--- a/chall02/krusthol.c
+++ b/chall02/krusthol.c
@@ -1,0 +1,51 @@
+#include <stdlib.h>
+
+void	double_hex_color(char *hex, char *rgb_hex_value, int value, int loc)
+{
+	rgb_hex_value[loc] = hex[value % 16];
+	rgb_hex_value[loc + 1] = hex[(value /= 16) % 16];
+}
+
+void	single_hex_color(char *hex, char *rgb_hex_value, int value, int loc)
+{
+	rgb_hex_value[loc] = '0';
+	rgb_hex_value[loc + 1] = hex[value];
+}
+
+char	*error_range(char *rgb_hex_value)
+{
+	rgb_hex_value[1] = 'e';
+	rgb_hex_value[2] = 'r';
+	rgb_hex_value[3] = 'r';
+	rgb_hex_value[4] = 'o';
+	rgb_hex_value[5] = 'r';
+	rgb_hex_value[6] = '!';
+	return (rgb_hex_value);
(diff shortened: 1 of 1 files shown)


krusthol:
And upon review I found that my output was wrong, I had to correct a critical bug. While I did, I also refactored the code, as detailed in the commit message. Basic idea is unchanged, but the code is simplified after the refactoring, with perhaps better style overall.

chall02/krusthol.c
```diff
@@ -0,0 +1,43 @@
...
+static char	*error_range(char *hex_value)
+{
+	hex_value[1] = 'e';
+	hex_value[2] = 'r';
+	hex_value[3] = 'r';
+	hex_value[4] = 'o';
+	hex_value[5] = 'r';
+	hex_value[6] = '!';
+	return (hex_value);
+}
```
jgengo, line 26: ahahah :)
nice, but no if there is an input error never send back this, return NULL is better for the other function that would use hv_rgb2hex to check if ret is null or no (like malloc do)

chall02/krusthol.c
```diff
@@ -0,0 +1,43 @@
...
+{
+	char	hex_table[16] = { "0123456789abcdef" };
+	char	*hex_value;
+
+	if (!(hex_value = (char *)malloc(sizeof(char) * 8)))
+		return (NULL);
+	hex_value[0] = '#';
+	hex_value[7] = '\0';
+	if ((r > 255 || g > 255 || b > 255) || (r < 0 || g < 0 || b < 0))
+		return (error_range(hex_value));
```
jgengo, line 38: would be better to check user input before allocating memories. But I see why you did it here.

jgengo (approved):
You come up with your genuine solution. I would suggest `sprintf();` but it's also good you come up with your own genuine solution. Much appreciate.

PR #90, opened 2020-03-30
ptuukkan's solution

ptuukkan:
This forces the value to between 0 and 255 to avoid user errors, then converts each color to hex. Uses static variable to avoid malloc.

### DIFF
--- a/chall02/ptuukkan.c
+++ b/chall02/ptuukkan.c
@@ -10,7 +10,7 @@
 /*                                                                            */
 /* ************************************************************************** */
 
-static char	itoc(int i)
+static char	ft_itoc(int i)
 {
 	if (i < 10)
 		return (i + '0');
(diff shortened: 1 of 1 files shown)


chall02/ptuukkan.c
```diff
@@ -0,0 +1,45 @@
...
+	else if (r < 0)
+		r = 0;
+	if (g > 255)
+		g = 255;
+	else if (g < 0)
+		g = 0;
+	if (b > 255)
+		b = 255;
+	else if (b < 0)
+		b = 0;
```
jgengo, line 35: it's nice you protected; cap the user inputs. 

a function should avoid changing the input and should return an error or null in this case if a user input isn't good.

Still, it's nice you came up with this solution. To avoid the if-forest and be a bit my DRY (don't repeat yourself) you could have written a function for that. `int cap_color(int color);`

jgengo (approved):

PR #91, opened 2020-03-30
Chall02 Solution

tdawson-42:
I use bitmasking with the bitwise AND operator and the value 0xff (255) to discard bits from the integer values so that r, g and b are all between 0-255, and then use bitshifting to move the bits to the correct position. Since the bitmask ensures rgb channel values are between 0-255 I do not check for input outside of that range since the subject doesn't specify required behaviour in such cases. Finally we format the hex value as a string starting with a '#' using sprintf.

### DIFF
--- a/chall02/tdawson.c
+++ b/chall02/tdawson.c
@@ -1,0 +1,15 @@
+#include <stdio.h>
+#include <unistd.h>
+
+char *hv_rgb2hex(int r, int g, int b)
+{
+	char *hex_str;
+
+	if (r < 0 || g < 0 || b < 0)
+		return NULL;
+
+	hex_str = malloc(8);
+	unsigned int hex = ((r & 0xff) << 16) + ((g & 0xff) << 8) + (b & 0xff);
+	sprintf(hex_str, "#%x", hex);
+	return hex_str;
+}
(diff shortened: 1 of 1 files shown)


chall02/tdawson.c
```diff
@@ -0,0 +1,13 @@
+#include <stdio.h>
+#include <stdlib.h>
+
+char *hv_rgb2hex(int r, int g, int b)
+{
+	char *hex_str;
+
+	if (!(hex_str = malloc(8)))
+		return NULL;
+	unsigned int hex = ((r & 0xff) << 16) + ((g & 0xff) << 8) + (b & 0xff);
```
jgengo, line 10: wuw :) bitwise!

where did you get it?

chall02/tdawson.c
```diff
@@ -0,0 +1,13 @@
...
+	sprintf(hex_str, "#%06x", hex);
```
jgengo, line 11: if you are using sprintf, you could have avoid bitwise 

```c
sprint(hex_str, "#%02x%02x%02x", r, g, b);
```

jgengo (approved):
really clear, and nice you used bitwise!

chall02/tdawson.c
Astridchan: I found it when doing research for FDF but I can't remember where, probably from Stack Overflow (where else 😁)

tdawson-42: I found it when doing research for my FDF but I can't remember where, probably Stack Overflow (where else 😁)

jgengo: Stack Overflow :) the developer toolbox!

PR #92, opened 2020-03-30
Create llahti solution to chall02

lapaset:
After checking that the function parameters are valid I allocate space for the string to be returned. Then I fill in the '#' and the colors sending the string and the parameters one by one to my int_to_hexa function and finally end the string with a null pointer.

### DIFF
--- a/chall02/llahti.c
+++ b/chall02/llahti.c
@@ -1,0 +1,31 @@
+#include <stdlib.h>
+
+char    *int_to_hexa(int nb, char *str, int i)
+{
+    char    *base = "0123456789abcdef";
+    int     len = 2;
+
+	while (len)
+	{
+		len--;
+		str[i + len] = base[nb % 16];
+		nb /= 16;
+	}
+	return (str);
+}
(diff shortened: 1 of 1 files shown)


chall02/llahti.c
```diff
@@ -0,0 +1,31 @@
...
+    int     len = 2;
+
+	while (len)
+	{
+		len--;
+		str[i + len] = base[nb % 16];
+		nb /= 16;
+	}
+	return (str);
+}
```
jgengo, line 15: really good!

chall02/llahti.c
```diff
@@ -0,0 +1,31 @@
...
+
+char     *hv_rgb2hex(int r, int g, int b)
+{
+    char    *str;
+
+    if (r < 0 || r > 255 || g < 0 || g > 255 || b < 0 || b > 255)
```
jgengo, line 21: it;s good you checked the user inputs!

jgengo (approved):

PR #94, opened 2020-03-30
Solution #02

tuommii:
Checking if values are valid and then converting those to hex with sprintf-function. No magic numbers.

### DIFF
--- a/chall02/mtuomine.c
+++ b/chall02/mtuomine.c
@@ -1,0 +1,35 @@
+#include <stdio.h>
+#include <stdlib.h>
+
+#define ARR_LEN 3
+#define HEX_LEN 7
+#define MIN_VALUE 0
+#define MAX_VALUE 255
+
+static int is_valid(int arr[ARR_LEN])
+{
+	size_t i = 0;
+	while (i < ARR_LEN)
+	{
+		if (arr[i] < MIN_VALUE || arr[i] > MAX_VALUE)
+			return 0;
+		i++;
+	}
+	return 1;
+}
+
+char *hv_rgb2hex(int r, int g, int b)
+{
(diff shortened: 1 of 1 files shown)


chall02/mtuomine.c
```diff
@@ -0,0 +1,35 @@
+#include <stdio.h>
+#include <stdlib.h>
+
+#define ARR_LEN 3
+#define HEX_LEN 7
+#define MIN_VALUE 0
+#define MAX_VALUE 255
```
jgengo, line 7: why not :) a bit overkill for such a small project.

chall02/mtuomine.c
```diff
@@ -0,0 +1,35 @@
...
+	int arr[ARR_LEN] = {r, g, b};
+
+	if (!is_valid(arr))
+		return NULL;
+
+	char *hex = (char *)malloc((sizeof(char) * HEX_LEN) + 1);
+	if (!hex)
+		return NULL;
+
+	sprintf(hex, "%s%02x%02x%02x", "#", r, g, b);
```
jgengo, line 32:
```c
sprintf(hex, "#%02x%02x%02x, r, g, b);
```

why did you add this const as an arg?

chall02/mtuomine.c
```diff
@@ -0,0 +1,35 @@
...
+	hex[HEX_LEN] = '\0';
```
jgengo, line 33: unnecessary, sprintf already add a trailing null

jgengo (commented):
good job!

chall02/mtuomine.c
tuommii: Easier to replace it or make it dynamic in the future, So my asnwer is maintainability 😂

PR #96, opened 2020-03-30
Just re-purposed libft functions 

RandommUser:
Hardcoded some things given the set string length. Used itoa_base to change the ints to hexadecimals.

### DIFF
--- a/chall02/phakakos.c
+++ b/chall02/phakakos.c
@@ -1,0 +1,94 @@
+/* ************************************************************************** */
+/*                                                                            */
+/*                                                        :::      ::::::::   */
+/*   phakakos.c                                         :+:      :+:    :+:   */
+/*                                                    +:+ +:+         +:+     */
+/*   By: phakakos <phakakos@student.hive.fi>        +#+  +:+       +#+        */
+/*                                                +#+#+#+#+#+   +#+           */
+/*   Created: 2020/02/10 17:26:08 by phakakos          #+#    #+#             */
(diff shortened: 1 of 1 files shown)


chall02/phakakos.c
```diff
@@ -0,0 +1,99 @@
...
+
+static int			cat_end(char *s1)
+{
+	int i;
+
+	i = 0;
+	while (s1[i])
+		i++;
+	return (i);
+}
```
jgengo, line 23: it's strlen?

`i` can't be negative, I would suggest size_t type

chall02/phakakos.c
```diff
@@ -0,0 +1,99 @@
...
+
+static char		base_conv(long value, int base)
+{
+	int	c;
+
+	c = value % base;
+	if (c > 9)
+		return ('a' + (c - 10));
+	return ('0' + c);
```
jgengo, line 51: smart!

chall02/phakakos.c
```diff
@@ -0,0 +1,99 @@
...
+	rtn[0] = '0';
+	itoa_conv(&rtn, value, base);
+	return (rtn);
+}
+
+char 			*hv_rgb2hex(int r, int g, int b)
+{
+	char *rtn;
+
+	if (r > -1 && r < 256 && g > -1 && g < 256 && b > -1 %% b < 256)
```
jgengo, line 86: good but would be better the other way around to split your error case handling and logic.

jgengo (approved):
good job!

PR #97, opened 2020-03-30
chall02 by sluhtala

sluhtala:
Hello,
First checking if the integers are in the range of 0 - 255. Then allocating spots for the string. Then I calculate individually the hex code for each spots. r divided by 16 and the r modulo 16. And if it is over 9 then return letter instead of number.

### DIFF
--- a/chall02/sluhtala.c
+++ b/chall02/sluhtala.c
@@ -1,0 +1,30 @@
+#include <stdlib.h>
+
+static	char base_16(int n)
+{
+	char c;
+
+	if (n < 10)
+		return (n + '0');
+	else
+		return (n - 10 + 'a');
+}
+
+char *hv_rgb2hex(int r, int g, int b)
+{
+	char *hex;
+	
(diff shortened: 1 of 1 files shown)


chall02/sluhtala.c
```diff
@@ -0,0 +1,30 @@
...
+		return (n + '0');
+	else
+		return (n - 10 + 'a');
+}
+
+char *hv_rgb2hex(int r, int g, int b)
+{
+	char *hex;
+	
+	if (r < 0 || g < 0 || b < 0 || r > 255 || g > 255 || b > 255)
```
jgengo, line 17: it's good you protected the user input

chall02/sluhtala.c
```diff
@@ -0,0 +1,30 @@
+#include <stdlib.h>
+
+static	char base_16(int n)
+{
+	char c;
+
+	if (n < 10)
+		return (n + '0');
+	else
+		return (n - 10 + 'a');
```
jgengo, line 10: smart!

jgengo (approved):
Good!

PR #98, opened 2020-03-30
chall02

ihwang:
Summary
1. Check if r, g and b are in range0-255.
2. Convert the given to char(hex) respectively.
3. Store the converted into Its correct position in 'ret'

### DIFF
--- a/chall02/ihwang.c
+++ b/chall02/ihwang.c
@@ -6,7 +6,7 @@
 /*   By: ihwang <marvin@42.fr>                      +#+  +:+       +#+        */
 /*                                                +#+#+#+#+#+   +#+           */
 /*   Created: 2020/03/30 14:13:25 by ihwang            #+#    #+#             */
-/*   Updated: 2020/03/30 14:45:01 by ihwang           ###   ########.fr       */
+/*   Updated: 2020/03/30 14:55:20 by ihwang           ###   ########.fr       */
 /*                                                                            */
 /* ************************************************************************** */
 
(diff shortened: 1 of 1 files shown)


chall02/ihwang.c
```diff
@@ -0,0 +1,63 @@
...
+	else if (cp == 12)
+		ret = 'c';
+	else if (cp == 13)
+		ret = 'd';
+	else if (cp == 14)
+		ret = 'e';
+	else if (cp == 15)
+		ret = 'f';
+	else
+		ret = cp + 48;
```
jgengo, line 36: This if-forest could be avoided by declaring a `char* base="0123456789abcdef"`

chall02/ihwang.c
```diff
@@ -0,0 +1,63 @@
...
+}
+
+char		*hv_rgb2hex(int r, int g, int b)
+{
+	char	*ret;
+
+	if (r > 255 || g > 255 || b > 255)
+		return (NULL);
+	if (r < 0 || g < 0 || b < 0)
+		return (NULL);
```
jgengo, line 53: It's good you check the user input.

chall02/ihwang.c
```diff
@@ -0,0 +1,63 @@
...
+	if (!(ret = (char*)malloc(sizeof(char) * 8)))
+		return (NULL);
+	bzero(ret, 8);
```
jgengo, line 56: good practice, but why? you knew you would fill ret anyway

jgengo (approved):
Good job! 

You should always try to avoid if-forest, and there is always a way to.

PR #99, opened 2020-03-30
chall-02 mdesta

mdesta:
Simply used sprintf to convert to hex, specifying two digits for each and return the resulting string

### DIFF
--- a/chall02/mdesta.c
+++ b/chall02/mdesta.c
@@ -1,0 +1,15 @@
+/* 
+** hv_rgb2hex - A function to convert colours in RGB
+** to Hexadecimal value
+*/
+#include <stdio.h>
+#include <stdlib.h>
+
+char *hv_rgb2hex(int r, int g, int b)
+{
+    char *hex;
+
+    hex = (char*)malloc(7 * sizeof(char) + 1);
+    sprintf(hex, "#%02x%02x%02x", r, g, b);
+    return (hex);
+}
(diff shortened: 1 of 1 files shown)


chall02/mdesta.c
```diff
@@ -0,0 +1,38 @@
...
+    if (r > g && b > r && b > 255)
+        return (0);   
+    if (r < g && g > b && g > 255)
+        return (0); 
+    return (1);
+}
+
+char *hv_rgb2hex(int r, int g, int b)
+{
+    if (validate_rgb(r, g, b))
```
jgengo, line 30: good!

chall02/mdesta.c
```diff
@@ -0,0 +1,38 @@
...
+    {
+        char *hex;
+        hex = (char*)malloc(7 * sizeof(char) + 1);
```
jgengo, line 33: not protected :(

chall02/mdesta.c
```diff
@@ -0,0 +1,38 @@
...
+        sprintf(hex, "#%02x%02x%02x", r, g, b);
```
jgengo, line 34: best usage of sprintf! nice job!

chall02/mdesta.c
jgengo, line 30: I think it would have been better to write:
```c
if (!(validate_rgb(r, g , b)
  return NULL;

char *hex;
```

it avoids to indent your whole code for just a check.

jgengo (approved):
Sadly, malloc isn't protected.
Sprintf really well used!

PR #100, opened 2020-03-30
chall02 submit

Rouhija:
1. Input validation
2. Malloc 8 bytes to buffer (hash, 6 digits and null byte)
3. Use sprintf to format leading hashtag and integers to two-digit hexes (sprintf null terminates the string)
4. Return the resulting 'ret' string

### DIFF
--- a/chall02/srouhe.c
+++ b/chall02/srouhe.c
@@ -1,0 +1,18 @@
+#include <stdio.h>
+#include <stdlib.h>
+
+char	*hv_rgb2hex(int r, int g, int b)
+{
+	char *ret;
+
+	if (r > 255 || r < 0 ||
+		g > 255 || g < 0 ||
+		b > 255 || b < 0)
+		return (NULL);
+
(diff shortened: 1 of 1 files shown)


chall02/srouhe.c
```diff
@@ -0,0 +1,18 @@
...
+#include <stdlib.h>
+
+char	*hv_rgb2hex(int r, int g, int b)
+{
+	char *ret;
+
+	if (r > 255 || r < 0 ||
+		g > 255 || g < 0 ||
+		b > 255 || b < 0)
+		return (NULL);
```
jgengo, line 11: good you protected the user inputs

chall02/srouhe.c
```diff
@@ -0,0 +1,18 @@
...
+
+	if (!(ret = (char *)malloc(sizeof(char) * 8)))
+		return (NULL);
```
jgengo, line 14: good!

jgengo (approved):

PR #101, opened 2020-03-30
Solution for chall02

AleXwern:
I could have used a static 7 char long string but since I'm doing graphics branch and I spend immense amount of time with colors, this could be useful and so I made the string with malloc to standardise it with the rest of the libft.

To the solution: We have 7 char long string and the index 0 is always set to '#'. After that we have 2 characters for R G and B respectively. At first the ft_puthex did all the work but I noticed some bugs with the character placement so I split it into 2 functions -> one handles the first one and the other the second. The position of the index we want to put characters into is tracked with static integer i that is fully handled in ft_puthex and when we need to place the first character in ft_firhex it's just added with +1 beforehand.
ft_puthex always puts something in the second slot but when we have ex. 15 which is F we needed something to notice that and in that case put 0 in the first slot so we dont have output like #fff but #0f0f0f. Thats where the ft_firhex comes in.

### DIFF
--- a/chall02/anystrom.c
+++ b/chall02/anystrom.c
@@ -1,0 +1,66 @@
+/* ************************************************************************** */
+/*                                                                            */
+/*                                                        :::      ::::::::   */
+/*   anystrom.c                                         :+:      :+:    :+:   */
+/*                                                    +:+ +:+         +:+     */
+/*   By: AleXwern <alex.nystrom5@gmail.com>         +#+  +:+       +#+        */
+/*                                                +#+#+#+#+#+   +#+           */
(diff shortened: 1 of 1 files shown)


chall02/anystrom.c
```diff
@@ -0,0 +1,66 @@
...
+	}
+	i++;
+	rgb[i] = c;
+}
+
+char	*hv_rgb2hex(int r, int g, int b)
+{
+	char	*rgb;
+
+	rgb = (char*)malloc(sizeof(char) * 7);
```
jgengo, line 60: You should always protect your malloc

jgengo (approved):
Good

PR #102, opened 2020-03-30
RGB to hexadecimal vkuikka

vkuikka:
First allocates the result string with the starting hash character and zeroes. Then passes addresses corresponding to color hexadecimal locations in the result string to another function that converts the color value to hexadecimal and puts that in the result string.

### DIFF
--- a/chall02/vkuikka.c
+++ b/chall02/vkuikka.c
@@ -17,7 +17,8 @@ char			*hv_rgb2hex(int r, int g, int b)
 		g < 0 || g > 255 ||
 		b < 0 || b > 255)
 		return (NULL);
-	res = strdup("#000000");
+	if (!(res = strdup("#000000")))
+		return (NULL);
 	ft_hex_to_str(res + 1, r);
 	ft_hex_to_str(res + 3, g);
 	ft_hex_to_str(res + 5, b);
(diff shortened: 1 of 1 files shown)


chall02/vkuikka.c
```diff
@@ -0,0 +1,25 @@
+#include <string.h>
+
+static void		ft_hex_to_str(char *str, int num)
+{
+	char *hex;
+	
+	hex = "0123456789abcdef";
+	str[1] = hex[num % 16];
+	str[0] = hex[num / 16 % 16];
+}
```
jgengo, line 10: super clean!

chall02/vkuikka.c
```diff
@@ -0,0 +1,25 @@
...
+
+char			*hv_rgb2hex(int r, int g, int b)
+{
+	char	*res;
+
+	if (r < 0 || r > 255 ||
+		g < 0 || g > 255 ||
+		b < 0 || b > 255 ||
+		!(res = strdup("#000000")))
+		return (NULL);
```
jgengo, line 20: smart!

I would suggest separate user inputs error cases handling and memory allocation.

jgengo (approved):
really good

PR #103, opened 2020-03-30
chall02

Rakutaa:
return null if r,g,b > 255 || r,g,b < 0
else
using light itoa_16 to convert rgb int parameters to a hex.
in the end catting "#" and RGB strings to a return value.


chall02/vtran.c
```diff
@@ -0,0 +1,27 @@
+#include <string.h>
+#include <stdlib.h>
+
+static char*    light_itoa_base(int nb, int base)
+{
+    char str[17];
+    char ret[3];
+
+    strcpy(str, "0123456789abcdef");
```
jgengo, line 9: you could have declared str directly since you don't have to respect the school norm :)

chall02/vtran.c
```diff
@@ -0,0 +1,27 @@
...
+    ret[0] = str[(nb / base) % base];
+    return (strdup(ret));
+}
+
+char* hv_rgb2hex(int r, int g, int b)
+{
+    char ret[8];
+
+    if (r > 255 || g > 255 || b > 255 || r < 0 || g < 0 || b < 0)
+        return (NULL);
```
jgengo, line 21: good!

jgengo (approved):
a bit special to see strdup in a ret.

PR #104, opened 2020-03-30
spentti chall02

saraarmiida:
I allocate space for the string and then use a modified ft_itoa_base to fill in the string with the given decimal numbers converted to hexadecimal numbers.

### DIFF
--- a/chall02/spentti.c
+++ b/chall02/spentti.c
@@ -1,0 +1,32 @@
+#include <stdlib.h>
+
+char	*ft_itoa_hex(unsigned long long n, char *str, int i)
+{
+	if (n < 16)
+	{
+		str[i - 1] = '0';
+		if (n == 0)
+			str[i] = '0';
+	}
+	while (n > 0)
+	{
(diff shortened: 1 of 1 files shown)


chall02/spentti.c
```diff
@@ -0,0 +1,32 @@
...
+{
+	if (n < 16)
+	{
+		str[i - 1] = '0';
+		if (n == 0)
+			str[i] = '0';
+	}
+	while (n > 0)
+	{
+		str[i] = (n % 16 < 10) ? '0' + n % 16 : n % 16 + 'a' - 10;
```
jgengo, line 13: wuw!

chall02/spentti.c
```diff
@@ -0,0 +1,32 @@
+#include <stdlib.h>
+
+char	*ft_itoa_hex(unsigned long long n, char *str, int i)
```
jgengo, line 3: This function could be written simpler, but it's good you came up with your own solution/converter

jgengo (approved):

PR #105, opened 2020-03-30
elindber chall02 solution

einoob:
My first solution used pretty basic itoa with the only difference that it would add the front zero if required. Everytime after calling itoa I used strjoin to join #->r->g->b respectively. That resulted in memory leaks due to strjoin.

This solution has some hard coded values to be able to use malloc only once. The string returned is always 7 characters long so we can always allocate the same amount of memory. Itoa won't need mallocing because we send the string and the place where to put the two numbers of the code. The program won't have any memory leaks when you free the string returned by hv_rgb2hexa() in the function that calls it.

### DIFF
--- a/chall02/elindber.c
+++ b/chall02/elindber.c
@@ -1,0 +1,46 @@
+/* ************************************************************************** */
+/*                                                                            */
+/*                                                        :::      ::::::::   */
+/*   elindber.c                                         :+:      :+:    :+:   */
+/*                                                    +:+ +:+         +:+     */
+/*   By: elindber <elindber@student.hive.fi>        +#+  +:+       +#+        */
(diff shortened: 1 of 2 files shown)


chall02/elindber.c
```diff
@@ -0,0 +1,46 @@
...
+		count++;
+	}
+	return (str);
+}
+
+char	*hv_rgb2hex(int r, int g, int b)
+{
+	char	*rgb_code;
+
+	if (r < 0 || g < 0 || b < 0 || r > 255 || g > 255 || b > 255)
```
jgengo, line 36: good!

jgengo (commented):
Good job!

Unfortunately I've to close your MR/PR; You pushed another file that shouldn't be here.

PR #106, opened 2020-03-30
hexadecimal

mashalink:
1.  Check if data is not less than 0 and not more than 255.
2. Create new clean string and string with hex value.
3.  Save # in new string, then simultaneously move 3 digits into hexadecimal format and save them. 
4. Return the result.

### DIFF
--- a/chall02/mlink.c
+++ b/chall02/mlink.c
@@ -1,0 +1,24 @@
+#include <stdlib.h>
+
+char *hv_rgb2hex(int r, int g, int b)
+{
+    char *str;
+    char *hex;
+
+    hex = "0123456789abcdef";
+    if (r < 0 || g < 0 || b < 0 || r > 255 || g > 255 || b > 255)
+        return (NULL);
+    str = (char*)malloc(sizeof(char) * 8);
+    str[7] = '\0';
+    str[0] = '#';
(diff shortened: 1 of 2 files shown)


chall02/mlink.c
```diff
@@ -0,0 +1,42 @@
...
+
+char *hv_rgb2hex(int r, int g, int b)
+{
+    char *str;
+    char *hex;
+
+    hex = "0123456789abcdef";
+    if (r < 0 || g < 0 || b < 0 || r > 255 || g > 255 || b > 255)
+        return (NULL);
+    str = (char*)malloc(sizeof(char) * 8);
```
jgengo, line 11: malloc not protected.

chall02/mlink.c
```diff
@@ -0,0 +1,42 @@
+#include <stdlib.h>
+
+char *hv_rgb2hex(int r, int g, int b)
+{
+    char *str;
+    char *hex;
+
+    hex = "0123456789abcdef";
+    if (r < 0 || g < 0 || b < 0 || r > 255 || g > 255 || b > 255)
```
jgengo, line 9: good!

jgengo (commented):
Good job, genuine solution.

Unfortunately, I've to close this PR, you send an extra file.

PR #107, opened 2020-03-30
Chall02

Epicurius:
Malloc 8 places in str.
Start of str place a '#' and at the end '\0'.
Using sprintf write into str hex value, moving them to apropriate place.
Return the finnished str.

### DIFF
--- a/chall02/nneronin.c
+++ b/chall02/nneronin.c
@@ -1,13 +1,19 @@
 #include <stdio.h>
 #include <stdlib.h>
 
+int	check(int value)
+{
+	return ((value < 0 || value > 255) ? 0 : 1);
+}
+
 char	*hv_rgb2hex(int r, int g, int b)
 {
 	char *str;
+	if (check(r) == 0 || check(g) == 0 || check(b) == 0)
+		return (NULL);
 	if (!(str = (char *)malloc(sizeof(char) * 8)))
 		return (NULL);
-	str[0] = '#';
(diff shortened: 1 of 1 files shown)


chall02/nneronin.c
```diff
@@ -0,0 +1,19 @@
...
+
+int	check(int value)
+{
+	return ((value < 0 || value > 255) ? 0 : 1);
+}
+
+char	*hv_rgb2hex(int r, int g, int b)
+{
+	char *str;
+	if (check(r) == 0 || check(g) == 0 || check(b) == 0)
```
jgengo, line 12: good! really good!

chall02/nneronin.c
```diff
@@ -0,0 +1,19 @@
...
+		return (NULL);
+	if (!(str = (char *)malloc(sizeof(char) * 8)))
+		return (NULL);
+	sprintf(&str[0], "#%x", (r & 0xFF) << 16 | (g & 0xFF) << 8 | (b & 0xFF));
```
jgengo, line 16: nice :) where did you find this bitwise?

```c
sprintf(str, "#%02x%02x%02x", r, g, b);
```
is clearer for everyone, but nice job with the bitwise

chall02/nneronin.c
```diff
@@ -0,0 +1,19 @@
...
+	str[7] == '\0';
```
jgengo, line 17: sprintf already add a trailing null, you don't need to do it.

jgengo (approved):
Good job! 
I love bitwise!

PR #108, opened 2020-03-30
Chall02

J0NY97:
for every color value i extract the hex values of the int and put it in string and return it.

### DIFF
--- a/chall02/jsalmi.c
+++ b/chall02/jsalmi.c
@@ -24,12 +24,12 @@ char *hv_rgb2hex(int r, int g, int b)
 
 	final = (char *)malloc(sizeof(char) * 8);
 	final[0] = '#';
-	final[7] = '\0';
 	final[1] = puthex((unsigned char)r / 16);
 	final[2] = puthex((unsigned char)r % 16);
 	final[3] = puthex((unsigned char)g / 16);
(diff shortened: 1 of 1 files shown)


chall02/jsalmi.c
```diff
@@ -0,0 +1,35 @@
...
+
+	if (nb >= 10)
+	{
+		nb == 10 ? c = 'a' : 0;
+		nb == 11 ? c = 'b' : 0;
+		nb == 12 ? c = 'c' : 0;
+		nb == 13 ? c = 'd' : 0;
+		nb == 14 ? c = 'e' : 0;
+		nb == 15 ? c = 'f' : 0;
+	}
```
jgengo, line 15: could have been done another way, it's not an if-forest but it's not really elegant. Plus, It will check all those conditions, not really well optimized.

chall02/jsalmi.c
```diff
@@ -0,0 +1,35 @@
...
+	else
+		c = nb + '0';
+	return (c);
+}
+
+char *hv_rgb2hex(int r, int g, int b)
+{
+	char *final;
+
+	final = (char *)malloc(sizeof(char) * 8);
```
jgengo, line 25: Don't forget to protect your malloc

jgengo (approved):

PR #111, opened 2020-03-30
chall02

avocadohooman:
First I store rgb values in in an int array.
Then I loop through each value and check if it is in the rgb range, if not return NULL.
If yes, allocate memory for HEX value + "#"
Then use sprintf to convert rgb values into hex with (%02x) and store in string. using "02" min width if digit is only one character add 0 in front of it .
Return str :)


chall02/gmolin.c
```diff
@@ -0,0 +1,33 @@
...
+	int		array[3] = {r, g, b};
+	int 	i;
+
+	i = 0;
+	while (i < 3)
+	{
+		if (array[i] < 0 || array[i] > 255)
+			return (NULL);
+		i++;
+	}
```
jgengo, line 28: quite funny way to check :) 

It would have been clearer to create a small function `int is_in_range(int c)` but I like the idea.

chall02/gmolin.c
```diff
@@ -0,0 +1,33 @@
...
+	if (!(str = (char*)malloc(sizeof(char) * 8)))
+		return (NULL);
+	sprintf(str, "#%02x%02x%02x", r, g, b);
```
jgengo, line 31: best usage of sprintf

jgengo (approved):

PR #114, opened 2020-03-30
chall02

shilena91:
My solution for chall02:
- Convert input numbers to hex (base 16):
**r** , **b**, **g** will become strings base 16 respectively
- After that join those strings together with prefix **#**

Good exercise for managing memory allocation!


jgengo:
You pushed once again all the files, PR refused.

shilena91:
> You pushed once again all the files, PR refused.

I saw in readme file I pull yesterday that "File to turn in" is sth _**.c_. Sorry my bad for misunderstanding.

Screenshot 2020-03-31 13 46 41

PR #115, opened 2020-03-30
Added chall02

MuSuareZ:
First what I do in the given function is to check if RBG values are respected to avoid misbehavior (0-255), I first allocate enough space for the string which will be returned in the end (1 hash and 6 hex values). For values from 0 to 9, I decided to use itoa() function on base 10 to convert these values from integers to chars and for the rest (10-15) I simply set their respective Hexadecimal values (a-f).
Once everything is calculated, ft_strjoin is called to join all the hexadecimal values to my first string previously allocated, following Red, Blue and Green order.

PS: My ft_strjoin() uses lots of my previously coded C functions so I had to bring most of them together(I'm so sorry, Norminette).

### DIFF
--- a/chall02/msuarez-.c
+++ b/chall02/msuarez-.c
@@ -1,0 +1,162 @@
+/* ************************************************************************** */
+/*                                                                            */
+/*                                                        :::      ::::::::   */
+/*   msuarez-.c                                         :+:      :+:    :+:   */
+/*                                                    +:+ +:+         +:+     */
+/*   By: msuarez- <msuarez-@student.hive.fi>        +#+  +:+       +#+        */
+/*                                                +#+#+#+#+#+   +#+           */
+/*   Created: 2020/03/30 15:20:29 by msuarez-          #+#    #+#             */
+/*   Updated: 2020/03/30 15:20:29 by msuarez-         ###   ########.fr       */
+/*                                                                            */
+/* ************************************************************************** */
+
+#include <stdlib.h>
+#include <string.h>
+
+static void		ft_bzero(void *s, size_t n)
+{
+	unsigned char	*p;
+	size_t			i;
+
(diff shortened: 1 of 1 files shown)


chall02/msuarez-.c
```diff
@@ -0,0 +1,162 @@
...
+	else if (value == 12)
+		tmp[0] = 'c';
+	else if (value == 13)
+		tmp[0] = 'd';
+	else if (value == 14)
+		tmp[0] = 'e';
+	else if (value == 15)
+		tmp[0] = 'f';
+	else
+		itoa(value, tmp, 10);
```
jgengo, line 107: You should always try to find a way not to use If-forest.

chall02/msuarez-.c
```diff
@@ -0,0 +1,162 @@
...
+	int		red_hex;
+	float	remainder;
+
+	red_hex = r / 16;
+	remainder = ((float)r / 16.00) - red_hex;
+	remainder = remainder * 16;
+	new = ft_strjoin(new, check_value(red_hex));
+	new = ft_strjoin(new, check_value(remainder));
+	return (new);
+}
```
jgengo, line 122: Having one function for each color isn't DRY at all.

chall02/msuarez-.c
```diff
@@ -0,0 +1,162 @@
...
+
+char			*hv_rgb2hex(int r, int g, int b)
+{
+	char	*new;
+
+	if (r < 0 || r > 255 || g < 0 || g > 255 || b < 0 || b > 255)
+		return (NULL);
+	new = ft_strnew(7);
```
jgengo, line 156: I think you should check new is not null

jgengo (approved):
You should try to use built-in function for those challenges, and also try to get familiar with the DRY concept to avoid repetition in your code.

PR #118, opened 2020-03-30
Chall02 (jnovotny)

jiricodes:
Firstly the validity of given numbers is checked (0 <= x <= 255).
Then memory is allocated for the output string, and first symbol is assigned to '#' and last to the end of the string ('\0'). Subsequently static function ft_hex is called for each of given numbers that writes their hex value in correct position of the string.

### DIFF
--- a/chall02/jnovotny.c
+++ b/chall02/jnovotny.c
@@ -1,0 +1,28 @@
+#include <stdlib.h>
+
+#define NOT_COLOR_INT(x) (x < 0 || x > 255)
+
+static char	*ft_hex(int value, int index, char *str)
+{
+	char *hex;
+	hex = "0123456789abcdef";
+	str[index + 1] = hex[value % 16];
+	value = value / 16;
+	str[index] = hex[value % 16];
+	return (str);
+}
+
+char		*hv_rgb2hex(int r, int g, int b)
+{
+	char *ret;
+
+	if (NOT_COLOR_INT(r) || NOT_COLOR_INT(g) || NOT_COLOR_INT(b))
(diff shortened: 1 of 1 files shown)


chall02/jnovotny.c
```diff
@@ -0,0 +1,28 @@
+#include <stdlib.h>
+
+#define NOT_COLOR_INT(x) (x < 0 || x > 255)
```
jgengo, line 3: I personally don't like functions inside define, but nice one!

chall02/jnovotny.c
```diff
@@ -0,0 +1,28 @@
...
+	return (str);
+}
+
+char		*hv_rgb2hex(int r, int g, int b)
+{
+	char *ret;
+
+	if (NOT_COLOR_INT(r) || NOT_COLOR_INT(g) || NOT_COLOR_INT(b))
+		return (NULL);
+	ret = (char *)malloc(8);
```
jgengo, line 21: not protected

chall02/jnovotny.c
```diff
@@ -0,0 +1,28 @@
...
+#define NOT_COLOR_INT(x) (x < 0 || x > 255)
+
+static char	*ft_hex(int value, int index, char *str)
+{
+	char *hex;
+	hex = "0123456789abcdef";
+	str[index + 1] = hex[value % 16];
+	value = value / 16;
+	str[index] = hex[value % 16];
+	return (str);
```
jgengo, line 12: nice

jgengo (approved):
good job!

PR #120, opened 2020-03-30
Chall02 - solution by hhuhtane

huhtasheikki:
### Chall02 - RGB to Hex string

Validation:
- Check that the values of **r**, **g** and **b** are between 0-255.

Make string for hex
- Malloc right amount of space (always 7 characters as '#' is the first character and each value uses two characters.
- Assign first ('#') and last (null character)
- Get hex values to string with int2hex funtion
-> skip already assigned characters
-> assign first the second character of color and then the first if there is still numbers left


jgengo:
Clear and clean code, 

Sadly I've to refuse this PR you pushed an extra file

PR #122, opened 2020-03-30
chall02 by mhasan

mhjony:
1. Error handling
2. Allocate the memory to store 7 characters.
3. Converted given integers into string using sprintf function and stored them in a char variable s1. 
4. Used strcat function to concate the strings
5. Return new string.


chall02/mhasan.c
```diff
@@ -0,0 +1,22 @@
...
+        return (NULL);
+    if (!(str = (char *)malloc(sizeof(char) * 8)))
+        return (NULL);
+    strcpy(str, "#");
+    sprintf(s1, "%02x", r);
+    strcat(str, s1);
+    sprintf(s1, "%02x", g);
+    strcat(str, s1);
+    sprintf(s1, "%02x", b);
+    strcat(str, s1);
```
jgengo, line 20: all of this could be written
```c
sprintf(s1, "#%02x%02x%02x", r, g, b);
```

jgengo (commented):
Unfortunately, I've to refuse this PR you pushed an extra file.

chall02/mhasan.c
mhjony: Could you please tell me what extra file I have pushed? I can't see anything extra in my local directory. I need to make sure it so that I won't do the same mistake in the future.

mhjony: Well, I think I found it.  It's Read.me file?

jgengo: yes

PR #124, opened 2020-03-31
chall02

tangxing1220:
I have done projects of FdF and Fractol. The knowledge of colors is helpful in this challenge.  
At the beginning,  I want to use ft_itoa_base() and ft_strjoin() in libft.  But these functions are general. Then I rewrite that functions for this challenge.
The idea of code is not complex.
- verify the input value of color, it will simplify the conversion int to str. 
- transfer int to hex string for every color component.
- assign string with 7 + 1 chars. 
- add every color component into string by strcat().


chall02/xtang.c
```diff
@@ -0,0 +1,52 @@
...
+}
+
+char		*hv_rgb2hex(int r, int g, int b)
+{
+	char *res;
+	char *str_r;
+	char *str_g;
+	char *str_b;
+
+	if (r < 0 || r > 255 || g < 0 || g > 255 || b < 0 || b > 255)
```
jgengo, line 39: good

jgengo (approved):
maybe a bit too much of malloc usage here, for a so small project but good job though!

PR #125, opened 2020-04-02
submit chall03 solution

tdawson-42:
My first code in Go :)

I first send a GET request to the specified URL and parse the ID and RGB values from the body message with Sscanf. Then I build a query string formatting the RGB values to hex with Sprintf. Finally I concatenate the original URL with the query string and send another GET request with the URL.

### DIFF
--- a/chall03/tdawson.go
+++ b/chall03/tdawson.go
@@ -7,8 +7,8 @@ import (
 	"net/http"
 )
 
-func httpGet() {
-	res, err := http.Get("https://chall03.hive.fi")
+func httpGet(url string) (bodyStr string) {
+	res, err := http.Get(url)
 	if err != nil {
 		log.Fatal(err)
 	}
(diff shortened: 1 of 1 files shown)


chall03/tdawson.go
```diff
@@ -0,0 +1,46 @@
...
+	"log"
+	"net/http"
+	"time"
+)
+
+func execTime(start time.Time) {
+	fmt.Printf("Finished in %dms\n", time.Since(start).Milliseconds())
+}
+
+func httpGet(url string) string {
```
jgengo, line 15: I would not suggest create a function httpGet, but use `http.Get()` inside your main scope. So if later you want to improve and use response Header or status code in your main code it's easier and cleaner.

chall03/tdawson.go
```diff
@@ -0,0 +1,46 @@
...
+	res.Body.Close()
+	if err != nil {
+		log.Fatal(err)
+	}
+	fmt.Printf("BODY:\t%s\n\n", string(body))
+	return string(body)
+}
+
+func buildQuery(body string) string {
+	var id, r, g, b int
```
jgengo, line 33: nice :)

chall03/tdawson.go
```diff
@@ -0,0 +1,46 @@
...
+	n, err := fmt.Sscanf(body, "id=%d,r=%d,g=%d,b=%d", &id, &r, &g, &b)
+	if err != nil || n != 4 {
+		log.Fatal(err)
+	}
+	return fmt.Sprintf("?id=%d&resp=%02x%02x%02x", id, r, g, b)
+}
+
+func main() {
+	defer execTime(time.Now())
```
jgengo, line 42: I wouldn't have thought about written this that way! nice!

jgengo (approved):

PR #126, opened 2020-04-02
add sadawi.go

salahadawi:
I convert the data i get from the request to https://chall03.hive.fi/ to a string. Then, I find the id and the 3 color integers I need from the string and convert the color integers to hex. I then concatenate everything to a string which i then use as my next get request url.

### DIFF
--- a/chall03/sadawi.go
+++ b/chall03/sadawi.go
@@ -1,0 +1,46 @@
+package main
+
+import (
+	"log"
+	"net/http"
+	"io/ioutil"
+	"strings"
+	"fmt"
+	"strconv"
+)
+
+func main() {
+	resp, err := http.Get("https://chall03.hive.fi/")
+	if err != nil {
+		log.Fatalln(err)
+	}
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		log.Fatalln(err)
+	}
+	log.Println(string(body))
+	s := strings.Split(string(body), ",")
+	id := s[0]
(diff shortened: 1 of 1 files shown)


chall03/sadawi.go
```diff
@@ -0,0 +1,41 @@
...
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		log.Fatalln(err)
+	}
+	log.Println(string(body))
+	s := strings.Split(string(body), ",")
+	id := s[0]
+	r, _ := strconv.Atoi(strings.Split(s[1], "=")[1])
+	g, _ := strconv.Atoi(strings.Split(s[2], "=")[1])
+	b, _ := strconv.Atoi(strings.Split(strings.Split(s[3], " ")[0], "=")[1])
```
jgengo, line 26: You should not ignore error in Go, and make sure the conversion here worked fine.

chall03/sadawi.go
```diff
@@ -0,0 +1,41 @@
...
+	rhex := fmt.Sprintf("%02x", r)
+	ghex := fmt.Sprintf("%02x", g)
+	bhex := fmt.Sprintf("%02x", b)
+	result :=  "https://chall03.hive.fi/?" + id + "&resp=" + rhex + ghex + bhex
```
jgengo, line 30: `+` concat is fine but I would recommend taking a look to fmt package and Sprintf()

jgengo (approved):

PR #127, opened 2020-04-02
chall03

osalmine:
First I get the values from the website, then I parse the values out of the string I got from the website, convert the rgb values to hex and get the response from the website with the id and the hex value.

### DIFF
--- a/chall03/osalmine.go
+++ b/chall03/osalmine.go
@@ -1,0 +1,51 @@
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"log"
+	"strings"
+	"strconv"
+)
+
+func get_rgb_id (s string) (int, int, int, int) {
+	str := strings.Split(s, " ")
+	vals := strings.Split(str[0], ",")
+	id, err := strconv.Atoi(strings.Split(vals[0], "=")[1]);
+	r, err := strconv.Atoi(strings.Split(vals[1], "=")[1]);
+	g, err := strconv.Atoi(strings.Split(vals[2], "=")[1]);
+	b, err := strconv.Atoi(strings.Split(vals[3], "=")[1]);
+	if err != nil {
+		log.Fatal(err)
+	}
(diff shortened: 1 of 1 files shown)


chall03/osalmine.go
```diff
@@ -0,0 +1,51 @@
...
+func get_rgb_id (s string) (int, int, int, int) {
+	str := strings.Split(s, " ")
+	vals := strings.Split(str[0], ",")
+	id, err := strconv.Atoi(strings.Split(vals[0], "=")[1]);
+	r, err := strconv.Atoi(strings.Split(vals[1], "=")[1]);
+	g, err := strconv.Atoi(strings.Split(vals[2], "=")[1]);
+	b, err := strconv.Atoi(strings.Split(vals[3], "=")[1]);
+	if err != nil {
+		log.Fatal(err)
+	}
```
jgengo, line 21: You should check `error != nil` right after each those lines.

chall03/osalmine.go
```diff
@@ -0,0 +1,51 @@
...
+	resp, err := http.Get(url)
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer resp.Body.Close()
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		log.Fatal(err)
+	}
+	s := fmt.Sprintf("%s", body)
```
jgengo, line 36: Can you explain why you did that? I don't really understand the purpose

jgengo (approved):

chall03/osalmine.go
osalmine: If you mean `s := fmt.Sprintf("%s", body)`, I was just converting the body that was a list of decimals I believe into a string. The only way I knew how to do this was to use sprintf. There is probably a better way of doing this but this was my first go code 😅

jgengo: You did a pretty good job.

I think `s := string(body)` is what you are looking for.

PR #128, opened 2020-04-02
chall03/npimenof

nikunicke:
get response from chall03.hive.fi
separate relevant values and convert numbers to int types
get hex value with printf
get request with url relevant params
print response

### DIFF
--- a/chall03/npimenof.go
+++ b/chall03/npimenof.go
@@ -23,9 +23,12 @@ func main() {
 	id := string(body)[0:strings.Index(string(body), ",")]
 	rgb := string(body)[strings.Index(string(body), ",")+1 : strings.Index(string(body), " ")]
 	rgb_separated := strings.Split(rgb, ",")
-	r, _ := strconv.Atoi(rgb_separated[0][2:])
-	g, _ := strconv.Atoi(rgb_separated[1][2:])
-	b, _ := strconv.Atoi(rgb_separated[2][2:])
+	r, err := strconv.Atoi(rgb_separated[0][2:])
+	g, err := strconv.Atoi(rgb_separated[1][2:])
+	b, err := strconv.Atoi(rgb_separated[2][2:])
+	if err != nil {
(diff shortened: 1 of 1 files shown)


chall03/npimenof.go
```diff
@@ -0,0 +1,39 @@
...
+	"fmt"
+	"io/ioutil"
+	"log"
+	"net/http"
+	"strconv"
+	"strings"
+)
+
+func main() {
+	client := http.Client{}
```
jgengo, line 13: It's unnecessary here, you can directly call the http.Get()

chall03/npimenof.go
```diff
@@ -0,0 +1,39 @@
...
+	if err != nil {
+		log.Fatal(err)
+	}
+	fmt.Println(string(body))
+	id := string(body)[0:strings.Index(string(body), ",")]
+	rgb := string(body)[strings.Index(string(body), ",")+1 : strings.Index(string(body), " ")]
+	rgb_separated := strings.Split(rgb, ",")
+	r, err := strconv.Atoi(rgb_separated[0][2:])
+	g, err := strconv.Atoi(rgb_separated[1][2:])
+	b, err := strconv.Atoi(rgb_separated[2][2:])
```
jgengo, line 28: you should check error right after each of them.

the 2 first could fail, and the last works, so you will keep running your code and have a problem at some point.

chall03/npimenof.go
```diff
@@ -0,0 +1,39 @@
...
+	"strings"
+)
+
+func main() {
+	client := http.Client{}
+	res, err := client.Get("https://chall03.hive.fi")
+	if err != nil {
+		log.Fatal(err)
+	}
+	body, err := ioutil.ReadAll(res.Body)
```
jgengo, line 18: you should close the res.Body() after using it. You can use defer or just write:

resp.Body().Close() after the error check.

jgengo (approved):

PR #129, opened 2020-04-02
Add solution alcohen.go to chall03

hiiri:
My first go program.

First get current timestamp and print it before and after sending GET requests. 
Parse response (in an inelegant way) from ex. 

> id=775,r=175,g=25,b=126 - Send...

 into separate variables: id, r, g, b. Use sprintf to convert rgb values to a hex formatted string and generate a new url using the id and hex string. Make a new get request using the formed url.

### DIFF
--- a/chall03/alcohen.go
+++ b/chall03/alcohen.go
@@ -10,6 +10,8 @@ import (
 	"log"
 )
 
+var g_time_start = time.Now()
+
 func make_get_request(url string) string {
 	resp, err := http.Get(url)
 	body, err := ioutil.ReadAll(resp.Body)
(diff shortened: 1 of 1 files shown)


chall03/alcohen.go
```diff
@@ -0,0 +1,56 @@
...
+	"time"
+)
+
+var gTimeStart = time.Now()
+
+func elapsedTime() time.Duration {
+	return time.Since(gTimeStart) / 1e6
+}
+
+func makeGetRequest(url string) string {
```
jgengo, line 19: http.Get() is already a net/http function that actually do that. I would suggest not to create a function for that, so if later you are asked to use a response Header or the response status code, you will not have to find a way to pass the parameters back to your main function.

chall03/alcohen.go
```diff
@@ -0,0 +1,56 @@
...
+	"net/http"
+	"strconv"
+	"strings"
+	"time"
+)
+
+var gTimeStart = time.Now()
+
+func elapsedTime() time.Duration {
+	return time.Since(gTimeStart) / 1e6
```
jgengo, line 16: `time.Since(gTimeStart).Milliseconds()`

chall03/alcohen.go
```diff
@@ -0,0 +1,56 @@
...
+	"fmt"
+	"io/ioutil"
+	"log"
+	"net/http"
+	"strconv"
+	"strings"
+	"time"
+)
+
+var gTimeStart = time.Now()
```
jgengo, line 13: I would avoid having a global variable for this

chall03/alcohen.go
```diff
@@ -0,0 +1,56 @@
...
+
+	body := makeGetRequest(url)
+	fmt.Printf("%dms - GET %s\n", elapsedTime(), url)
+	fmt.Printf("\tanswer:%s\n", body)
+
+	s := strings.Split(body, "=")
+	id := strings.Split(s[1], ",")[0]
+	r, _ := strconv.Atoi(strings.Split(s[2], ",")[0])
+	g, _ := strconv.Atoi(strings.Split(s[3], ",")[0])
+	b, _ := strconv.Atoi(strings.Split(s[4], " ")[0])
```
jgengo, line 45: you should not ignore errors here.

jgengo (approved):

PR #130, opened 2020-04-02
Solution #03

tuommii:
Created `http.Client` with timeout and extracted interesting values from first response and with those i created query string manually (GET-method instead of POST) for second request.

### DIFF
--- a/chall03/mtuomine.go
+++ b/chall03/mtuomine.go
@@ -1,0 +1,59 @@
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
+	"log"
+	"net/http"
+	"regexp"
+	"strconv"
+	"strings"
+	"time"
+)
+
+const baseURL = "https://chall03.hive.fi/"
+
+func main() {
+	client := http.Client{Timeout: time.Second * 1}
+
+	// First request
+	res, err := client.Get(baseURL)
+	checkErr(err)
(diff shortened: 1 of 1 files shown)


chall03/mtuomine.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+const baseURL = "https://chall03.hive.fi/"
+
+func main() {
+	client := http.Client{Timeout: time.Second * 1}
+
+	// First request
+	res, err := client.Get(baseURL)
+	checkErr(err)
```
jgengo, line 21: check error in Go can be repetitive, and breaking the DRY concept, but you should avoid to create function like this to handle your errors.

Some error should panic/quit, some should just raise a warning or print debug.

chall03/mtuomine.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+func checkErr(err error) {
+	if err != nil {
+		log.Fatal(err)
+	}
+}
+
+func parseNumbers(str string) []string {
+	re := regexp.MustCompile("[0-9]+")
```
jgengo, line 41: You should avoid regex for such a project, here a nice post you could read: https://softwareengineering.stackexchange.com/a/113243

jgengo (approved):

PR #131, opened 2020-04-02
HTTP RGB to Hexadecimal

vossr:
http get, read html body, cut away extra characters, make hex from rgb numbers, make url from id and hex, http get url and print result.

### DIFF
--- a/chall03/rpehkone.go
+++ b/chall03/rpehkone.go
@@ -1,0 +1,26 @@
+package main
+
+import (
+	"io/ioutil"
+	"fmt"
+	"strings"
+	"net/http"
+	"strconv"
+)
+
+func main() {
+	res, err := http.Get("https://chall03.hive.fi/")
+	body, err := ioutil.ReadAll(res.Body)
+	tmp := strings.Split(string(body), " ")
+	str := strings.Split(tmp[0], ",")
+	id, err := strconv.Atoi(str[0][3:len(str[0])])
+	r, err := strconv.Atoi(str[1][2:len(str[1])])
+	g, err := strconv.Atoi(str[2][2:len(str[2])])
+	b, err := strconv.Atoi(str[3][2:len(str[3])])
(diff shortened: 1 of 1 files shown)


chall03/rpehkone.go
```diff
@@ -0,0 +1,26 @@
...
+import (
+	"io/ioutil"
+	"fmt"
+	"strings"
+	"net/http"
+	"strconv"
+)
+
+func main() {
+	res, err := http.Get("https://chall03.hive.fi/")
```
jgengo, line 12: You should check `err != nil` after each function that has an error type returned.

chall03/rpehkone.go
```diff
@@ -0,0 +1,26 @@
...
+	body, err := ioutil.ReadAll(res.Body)
+	tmp := strings.Split(string(body), " ")
+	str := strings.Split(tmp[0], ",")
+	id, err := strconv.Atoi(str[0][3:len(str[0])])
+	r, err := strconv.Atoi(str[1][2:len(str[1])])
+	g, err := strconv.Atoi(str[2][2:len(str[2])])
+	b, err := strconv.Atoi(str[3][2:len(str[3])])
+	var rgb string = fmt.Sprintf("%02x%02x%02x", r, g, b);
+	var url string = "https://chall03.hive.fi/?id=" + strconv.Itoa(id) + "&resp=" + rgb
```
jgengo, line 21: url := fmt.Sprintf("") 

I suggest you take a look to fmt package there are pretty handy functions you could use to avoid the `+` concatenation.

chall03/rpehkone.go
```diff
@@ -0,0 +1,26 @@
...
+	res2, err := http.Get(url)
+	body2, err := ioutil.ReadAll(res2.Body)
+	fmt.Println(string(body2))
+	_ = err;
```
jgengo, line 25: You should never do that. Error case should be handled.

jgengo (approved):

PR #132, opened 2020-04-02
chall03

vkuikka:
Gets response from given address and converts the color values to hexadecimal. Then sends the hexadecimal color value back to the same address. Every time something is sent, the address, response and status is printed.

### DIFF
--- a/chall03/vkuikka.go
+++ b/chall03/vkuikka.go
@@ -1,0 +1,52 @@
+package main
+
+import (
+	"fmt"
+	"net/http"
+	"strings"
+	"strconv"
+	"io/ioutil"
+)
+
+func	getData(addr string) string {
+	fmt.Println(addr)
+	response, err := http.Get(addr)
+	if err != nil {
+		return ("error")
+	}
+	print(string(response.StatusCode) + response.Status + "\n")
+	responseData, err := ioutil.ReadAll(response.Body)
+	if err != nil {
+		return ("error")
+	}
+	responseString := string(responseData)
+	fmt.Println("answer: " + responseString)
+	return responseString
+}
+
+func	sendResponse(response string, addr string) int {
+	values_str := strings.Split(response, " ")[0]
+	values := strings.Split(values_str, ",")
+	r, _ := strconv.Atoi(strings.Split(values[1], "=")[1])
+	g, _ := strconv.Atoi(strings.Split(values[2], "=")[1])
+	b, _ := strconv.Atoi(strings.Split(values[3], "=")[1])
+	hex := fmt.Sprintf("%02x", r)
+	hex += fmt.Sprintf("%02x", g)
+	hex += fmt.Sprintf("%02x", b)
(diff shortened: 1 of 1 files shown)


chall03/vkuikka.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"fmt"
+	"net/http"
+	"strings"
+	"strconv"
+	"io/ioutil"
+)
+
+func	getData(addr string) string {
```
jgengo, line 11: I would not recommend creating another function to Get Data while you already have http.Get() for that.

Plus, if one day you have to parse more information return by http.Get() into your main code you will be forced to find a way to send more parameters, and it will start to make your code less clean.

chall03/vkuikka.go
```diff
@@ -0,0 +1,50 @@
...
+	fmt.Println("answer: " + responseString)
+	return responseString
+}
+
+func	sendResponse(response string, addr string) bool {
+	values_str := strings.Split(response, " ")[0]
+	values := strings.Split(values_str, ",")
+	r, _ := strconv.Atoi(strings.Split(values[1], "=")[1])
+	g, _ := strconv.Atoi(strings.Split(values[2], "=")[1])
+	b, _ := strconv.Atoi(strings.Split(values[3], "=")[1])
```
jgengo, line 32: You should avoid as much as possible to ignore error return in Go, and always check it. 

It can be repetitive, but it's better to check that the convert worked for example here.

chall03/vkuikka.go
```diff
@@ -0,0 +1,50 @@
...
+func	getData(addr string) string {
+	fmt.Println(addr)
+	response, err := http.Get(addr)
+	if err != nil {
+		return ("error")
+	}
+	print(string(response.StatusCode) + response.Status + "\n")
+	responseData, err := ioutil.ReadAll(response.Body)
+	if err != nil {
+		return ("error")
```
jgengo, line 20: in Go you can return 2 variables (or more)

This function should be re-written to have (string, error) return variables then here you could have written:

```go
return nil, errors.New("can't read the body")
```

chall03/vkuikka.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+func	sendResponse(response string, addr string) bool {
+	values_str := strings.Split(response, " ")[0]
+	values := strings.Split(values_str, ",")
+	r, _ := strconv.Atoi(strings.Split(values[1], "=")[1])
+	g, _ := strconv.Atoi(strings.Split(values[2], "=")[1])
+	b, _ := strconv.Atoi(strings.Split(values[3], "=")[1])
+	hex := fmt.Sprintf("%02x%02x%02x", r, g, b)
+	response = getData(addr + "?" + values[0] + "&resp=" + hex)
```
jgengo, line 34: I would suggest 

fmt.Sprintf() instead of concatenation.

jgengo (approved):
Good job!

PR #133, opened 2020-04-02
Add krusthol's solution to HTTP RGB to Hexadecimal

krusthol:
Approach:
First Golang code with a very beginner level style. Thus, it is broken to several functions, to make some sense out of it. The resulting code mass is large, unelegant. Breaking DRY concept with the printing as well several times. But, the end result worked during development and tested to be ok, and the print output was able to match the suggested output format of the challenge examples.

Code: 
HTTP GET request is made to the challenge address. Parsing is done using regexp functions, strconv.Atoi, strings.Split and for each -iteration for the values. Code assumes the body and value order stays stable and similiar as when the code was developed. Finally, another HTTP GET request is made with the hexadecimal & id values. The answer to this request is then printed to the console. Console prints are done using fmt.Printf and log.Println, so that milliseconds since execution begun is shown as prefix.

### DIFF
--- a/chall03/krusthol.go
+++ b/chall03/krusthol.go
@@ -1,0 +1,87 @@
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
+	"log"
+	"net/http"
+	"regexp"
+	"strconv"
+	"strings"
+	"time"
+)
+
+func get_chall(address string) (id_rgb_str string) {
+	fmt.Printf("%d", time.Since(StartTime).Milliseconds())
+	log.Println("ms - GET ", address)
+	chall_req, err := http.Get(address)
+	if err != nil {
+		log.Fatal(err)
+	}
+	raw_id_rgb, err := ioutil.ReadAll(chall_req.Body)
+	if err != nil {
+		log.Fatal(err)
+	}
+	chall_req.Body.Close()
+	id_rgb_str = fmt.Sprintf("%s", raw_id_rgb)
+	fmt.Printf("%d", time.Since(StartTime).Milliseconds())
+	log.Println("ms - GET ", address)
+	log.Println("\tanswer: ", id_rgb_str)
+	return
+}
+
+func regexp_parse_before_slash(chall_body string) string {
+	halved_string := strings.Split(chall_body, "-")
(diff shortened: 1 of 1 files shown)


krusthol:
Last minute refactor:
I had ideas how to improve DRY of my code, and implemented some simple refactoring for that.

chall03/krusthol.go
```diff
@@ -0,0 +1,70 @@
...
+)
+
+func log_prefixed(msg string) {
+	fmt.Printf("%d", time.Since(StartTime).Milliseconds())
+	log.Println("ms - ", msg)
+}
+
+func get_request(address string) (id_rgb_str string) {
+	log_prefixed("GET " + address)
+	chall_req, err := http.Get(address)
```
jgengo, line 21: Golang use CamelCase variable name type, not snake one

`challReq, err :=`

chall03/krusthol.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+	chall_req.Body.Close()
+	id_rgb_str = fmt.Sprintf("%s", raw_id_rgb)
+	log_prefixed("GET " + address + "\n\tanswer: " + id_rgb_str)
+	return
+}
+
+func regexp_parse_before_slash(chall_body string) string {
+	halved_string := strings.Split(chall_body, "-")
+	reg, err := regexp.Compile("[^0-9=]+")
```
jgengo, line 37: For this exercise I would recommend using Sscanf instead of regex: https://softwareengineering.stackexchange.com/a/113243

chall03/krusthol.go
```diff
@@ -0,0 +1,70 @@
...
+	"strings"
+	"time"
+)
+
+func log_prefixed(msg string) {
+	fmt.Printf("%d", time.Since(StartTime).Milliseconds())
+	log.Println("ms - ", msg)
+}
+
+func get_request(address string) (id_rgb_str string) {
```
jgengo, line 19: I would not recommend using a function to do the http request, because if you need to use, one day, a response Header or StatusCode in your main scope, you will need to return some parameters, and start to make the code less cleaner.

jgengo (approved):
Good job!

PR #134, opened 2020-04-02
wkorande chall03

wehard:
- Send a GET request
- Split the response at space
- Split the resulting strings at comma
- Split results at equals sign
- Get values from resulting at index 1
- Convert rgb values to hex
- Construct POST request and build query string using the parsed data
- Send the request
- Profit!

### DIFF
--- a/chall03/wkorande.go
+++ b/chall03/wkorande.go
@@ -4,11 +4,12 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"os"
 	"strconv"
 	"strings"
 )
 
-func convertRGB(rs, gs, bs string) string {
+func convertRGBToHex(rs, gs, bs string) string {
 	r, err := strconv.Atoi(rs)
 	if err != nil {
 		fmt.Println(err)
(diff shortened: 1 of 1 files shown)


chall03/wkorande.go
```diff
@@ -0,0 +1,64 @@
...
+	"net/http"
+	"os"
+	"strconv"
+	"strings"
+)
+
+func convertRGBToHex(rs, gs, bs string) string {
+	r, err := strconv.Atoi(rs)
+	if err != nil {
+		fmt.Println(err)
```
jgengo, line 15: I think the convert here is too important to be only a Println. It would have been better to use 
```go
log.Fatalln()
```
to exit in case there is a problem.

chall03/wkorande.go
```diff
@@ -0,0 +1,64 @@
...
+		os.Exit(1)
+	}
+	bd := strings.Split(string(getBody), " ")[0]
+	d := strings.Split(bd, ",")
+	id := strings.Split(d[0], "=")[1]
+	r := strings.Split(d[1], "=")[1]
+	g := strings.Split(d[2], "=")[1]
+	b := strings.Split(d[3], "=")[1]
+	hex := convertRGBToHex(r, g, b)
```
jgengo, line 47: This could fail if I change the body response. I would recommend using Sscanf here instead of Split().

chall03/wkorande.go
```diff
@@ -0,0 +1,64 @@
...
+	req, err := http.NewRequest("POST", "https://chall03.hive.fi", nil)
+	q := req.URL.Query()
+	q.Add("id", id)
+	q.Add("resp", hex)
+	req.URL.RawQuery = q.Encode()
+
+	postResponse, err := http.DefaultClient.Do(req)
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
```
jgengo, line 58: log.Fatalln(err) suits better in this case while log.Fatalln is a Println with os.Exit(1) after usage.

jgengo (approved):

PR #135, opened 2020-04-02
ptuukkan's solution

ptuukkan:
First HTTP GET request to the server, get body as string, then parse RGB and ID values from it. Then check if the RGB values are valid, convert them to a hex string and send a new GET request containing the ID and HEX strings. Finally we print the GET request's response body.

Thanks for the really fun challenge!

### DIFF
--- a/chall03/ptuukkan.go
+++ b/chall03/ptuukkan.go
@@ -1,0 +1,78 @@
+package main
+
+import (
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"net/http"
+	"strconv"
+	"strings"
+)
+
+func parseResponse(response string) (map[string]int, string) {
+	splitted := strings.Split(response, "-")
+	values := strings.Split(strings.Trim(splitted[0], " "), ",")
+	valuemap := make(map[string]int)
+	var id string = ""
+	for _, value := range values {
+		elem := strings.Split(value, "=")
+		if elem[0] == "id" {
+			id = elem[1]
+		} else {
+			valuemap[elem[0]], _ = strconv.Atoi(elem[1])
+		}
+	}
+	return valuemap, id
+}
+
+func validColors(values map[string]int) bool {
+	for _, value := range values {
+		if value > 255 || value < 0 {
+			return false
+		}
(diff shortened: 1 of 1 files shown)


chall03/ptuukkan.go
```diff
@@ -0,0 +1,78 @@
...
+	splitted := strings.Split(response, "-")
+	values := strings.Split(strings.Trim(splitted[0], " "), ",")
+	valuemap := make(map[string]int)
+	var id string = ""
+	for _, value := range values {
+		elem := strings.Split(value, "=")
+		if elem[0] == "id" {
+			id = elem[1]
+		} else {
+			valuemap[elem[0]], _ = strconv.Atoi(elem[1])
```
jgengo, line 23: You shouldn't ignore the error return.

chall03/ptuukkan.go
```diff
@@ -0,0 +1,78 @@
...
+		}
+	}
+	return valuemap, id
+}
+
+func validColors(values map[string]int) bool {
```
jgengo, line 29: nice function!

chall03/ptuukkan.go
```diff
@@ -0,0 +1,78 @@
...
+		if value > 255 || value < 0 {
+			return false
+		}
+	}
+	return true
+}
+
+func convertHex(values map[string]int) (string, error) {
+	if !validColors(values) {
+		return "", errors.New("invalid rgb values")
```
jgengo, line 40: even used `errors.New()`, nice!

chall03/ptuukkan.go
```diff
@@ -0,0 +1,78 @@
...
+	if err != nil {
+		log.Fatal(err)
+		return "", err
+	}
+	data, err := ioutil.ReadAll(response.Body)
+	if err != nil {
+		log.Fatal(err)
+		return "", err
+	}
+	response.Body.Close()
```
jgengo, line 57: nice!

jgengo (approved):
Good code!

That could have been nice to implement `time` to have more detail on how fast is the requests and whole execution time of the function.

PR #136, opened 2020-04-02
GoHex

kimsappi:
I first request the page with http.Get and then parse it with strings.Split and strconv.Atoi (trusting that they are numbers). I convert the parsed RGB values to hex with fmt.Sprintf and then make a URL with the parsed ID and convertex hex string. Finally I print out the HTTP response body and status code, assuming there haven't been errors with the HTTP requests and the bodies have been readable.

### DIFF
--- a/chall03/ksappi.go
+++ b/chall03/ksappi.go
@@ -1,0 +1,37 @@
+package main
+
+import (
+  "fmt"
+  "net/http"
+  "io/ioutil"
+  "strings"
+  "strconv"
+)
+
+func main() {
+  resp, err := http.Get("https://chall03.hive.fi/")
+  if err != nil {
+    return
+  }
+  body, err := ioutil.ReadAll(resp.Body)
+  if err != nil {
+    return
+  }
+  splitted := strings.Split(string(body), ",")
+
+  red, err := strconv.Atoi(strings.Split(splitted[1], "=")[1])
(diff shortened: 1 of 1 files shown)


chall03/ksappi.go
```diff
@@ -0,0 +1,37 @@
...
+  "strings"
+  "strconv"
+)
+
+func main() {
+  resp, err := http.Get("https://chall03.hive.fi/")
+  if err != nil {
+    return
+  }
+  body, err := ioutil.ReadAll(resp.Body)
```
jgengo, line 16: you should add one line above a defer to close `resp.Body` or after this scope.

chall03/ksappi.go
```diff
@@ -0,0 +1,37 @@
...
+  if err != nil {
+    return
+  }
+  splitted := strings.Split(string(body), ",")
+
+  red, err := strconv.Atoi(strings.Split(splitted[1], "=")[1])
+  green, err := strconv.Atoi(strings.Split(splitted[2], "=")[1])
+  blue, err := strconv.Atoi(strings.Split(strings.Split(splitted[3], "=")[1], " ")[0])
```
jgengo, line 24: err should be checked after usage.

and in Go if you are not using a return value you can use the **blank identifier** `_`

e.g 

```go
red, _ := strconv.Atoi()
```

jgengo (approved):
Good job!

PR #137, opened 2020-04-02
Oh god, why is every variable named differently to C

RandommUser:
First get the response. Make it into a string. Then I parse it to get the id, and color values with a very complicated "while" loop. Then I convert the 3 color values to hexadecimals. I append the strings together to get the color code. Then I combine the base url and the id and color value. Then I make the second request and print out the http response.

### DIFF
--- a/chall03/phakakos.go
+++ b/chall03/phakakos.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"log"
 	"strconv"
+	"time"
 )
 
 func hexaConv(num int) string {
@@ -32,6 +33,8 @@ func hexaConv(num int) string {
 
 
 func main() {
+	start := time.Now()
+	fmt.Printf("Making the request\n")
 	resp, err := http.Get("https://chall03.hive.fi/")
 	if err != nil {
     log.Fatal(err)
@@ -43,7 +46,8 @@ func main() {
 			log.Fatal(err)
 		}
 		bodyString := string(bodyBytes)
-		fmt.Printf(bodyString)
+		//t := time.Now()
+		fmt.Printf("%d MS | %s\n", time.Now().Sub(start).Milliseconds(), bodyString)
 		i, id, red, green, blue, valNum, reading := 0, 0, 0, 0, 0, 0, 0
 		for valNum != 4 {
 			if (bodyString[i] == '='){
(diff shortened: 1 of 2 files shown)


chall03/phakakos.go
```diff
@@ -0,0 +1,113 @@
...
+		b = rune('a' + u - 10)
+	} else {
+		b = rune('0' + u)
+	}
+	if (y > 9){
+		a = rune('a' + y - 10)
+	} else {
+		a = rune('0' + y)
+	}
+	ret = string(a) + string(b)
```
jgengo, line 29: I would suggest using:

```go
fmt.Sprintf()
```

chall03/phakakos.go
```diff
@@ -0,0 +1,113 @@
...
+
+
+func main() {
+	start := time.Now()
+	fmt.Printf("Making the request\n")
+	resp, err := http.Get("https://chall03.hive.fi/")
+	if err != nil {
+    log.Fatal(err)
+    }
+	defer resp.Body.Close()
```
jgengo, line 42: nice usage of defer.

chall03/phakakos.go
```diff
@@ -0,0 +1,113 @@
...
+		i, id, red, green, blue, valNum, reading := 0, 0, 0, 0, 0, 0, 0
+		for valNum != 4 {
+			if (bodyString[i] == '='){
+				reading = 1
+			} else if (bodyString[i] == ',' || bodyString[i] == ' '){
+				reading = 0
+				valNum++
+			}else{
+				if (reading == 1 && valNum == 0){
+					if (id > 0){
```
jgengo, line 60: This long part makes your code a little bit less comprehensive.

chall03/phakakos.go
```diff
@@ -0,0 +1,113 @@
...
+						}
+					blue += int(bodyString[i] - '0')
+				}
+			}
+			i++
+		}
+		fmt.Printf("%d MS | id %d red %d green %d blue %d\n", time.Now().Sub(start).Milliseconds(), id, red, green, blue)
+		
+		cRed, cGreen, cBlue := hexaConv(red), hexaConv(green), hexaConv(blue)
+		fColor := cRed + cGreen + cBlue
```
jgengo, line 92: fmt.Sprintf() would be better

jgengo (commented):
Good job!

chall03/phakakos.go
RandommUser: This was my first Golang project so I wanted to focus on figuring out how the strings manipulation works in the language, even it if might be convoluted as hell.

jgengo: no worries your solution is fine. And Actually, it sometime make sense to use that instead of importing the package fmt just for that.

In this package, I would avoid using fmt.Sprintf for example: 
```go 
// Package twofer is to share something with someone
package twofer

// ShareWith is doing something uncommentable :D
func ShareWith(name string) string {
	if name == "" {
		name = "you"
	}
	return "One for " + name + ", one for me." // I could have use Sprintf but would like to avoid useless import fmt
}
```

PR #138, opened 2020-04-02
Chall03

avocadohooman:
First little program  in Go, I had like 50 tabs open. But nice language, however, don't like the fmt.Println syntax (the fmt. part) but I LOVE ":=" and the possibility to handle errors in the same breath as declaring a variable (a, err) <3

First I send a GET request to  https://chall03.hive.fi/
Slice the body several times so I can get the id and rgb values
Then convert rgb to string with strconv.Atoi 
Then create HEX value using sprintf 
Then second GET request to https://chall03.hive.fi/?id={id}&resp={hex}

Fingers crossed it is fast enough!

### DIFF
--- a/chall03/gmolin.go
+++ b/chall03/gmolin.go
@@ -1,0 +1,1 @@
+package main
--- a/chall03/gmolin.go
+++ b/chall03/gmolin.go
@@ -8,7 +8,7 @@ import (
 	"strings"
 )
 
-func http_post_hex(id string, hex string)string{
+func http_get_hex(id string, hex string)string{
 	result, err := http.Get("https://chall03.hive.fi/?id="+id+"&resp="+hex)
 	if err != nil {
 		fmt.Println(err)
@@ -46,6 +46,6 @@ func main(){
 	g, _ := strconv.Atoi(g_array[1])
 	b, _ := strconv.Atoi(b_array_only_b[0])
 	hex := fmt.Sprintf("%02x%02x%02x", r, g, b)
-	body := http_post_hex(id, hex)
+	body := http_get_hex(id, hex)
 	fmt.Println(body)
 }
--- a/chall03/gmolin.go
+++ b/chall03/gmolin.go
@@ -1,1 +1,48 @@
 package main
+
+import (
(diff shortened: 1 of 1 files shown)


chall03/gmolin.go
```diff
@@ -0,0 +1,51 @@
...
+
+func http_get_id_rgb(www string)[]string{
+	result, err := http.Get(www)
+	if err != nil {
+		fmt.Println(err)
+	}
+	body, err := ioutil.ReadAll(result.Body)
+	if err != nil{
+		fmt.Println(err)
+	}
```
jgengo, line 31: Somehow, your both function are pretty doing the same thing, I would recommend not creating another function to do that but let the "main" logic into your main.

If tomorrow you are asked to use the `result` variable, you will have to pass/return parameters when it could be directly used in your main scope if it's declared there. (You may be asked to check a `resp.StatusCode`, for example, or `resp.Header()`

chall03/gmolin.go
```diff
@@ -0,0 +1,51 @@
...
+	fmt.Println(sliced_body)
+	id_array := strings.Split(sliced_body[0], "=")
+	r_array := strings.Split(sliced_body[1], "=")
+	g_array := strings.Split(sliced_body[2], "=")
+	b_array_all := strings.Split(sliced_body[3], "=")
+	b_array_only_b := strings.Split(b_array_all[1], " ")
+	id := id_array[1]
+	r, _ := strconv.Atoi(r_array[1])
+	g, _ := strconv.Atoi(g_array[1])
+	b, _ := strconv.Atoi(b_array_only_b[0])
```
jgengo, line 47: if the body response isn't like it should you will probably raise errors and your `strconv.Atoi` errors are ignored. 
in Golang we avoid as much as possible to ignore errors, I know it can super repetitive code but error handling is important.

jgengo (approved):
Good job!

Yes, at first, I was thinking the same `log.Println()` or `fmt.Println` everywhere and error handling 
```go
if err != nil {}
```
but actually now realised how much it's important and also good to know which package is used.

PR #139, opened 2020-04-02
my chall03 solution

shilena91:
- Use http.GET to make GET request from URL
- Use Sscanf to get id, r, b, g
- Use Sprintf to get new URL
- Make GET request from new URL

I hope this time I pulled and pushed correctly T_T.


chall03/hopham.go
```diff
@@ -0,0 +1,48 @@
...
+		log.Fatalln(err)
+	}
+	content, err := ioutil.ReadAll(result.Body)
+	if err != nil {
+		log.Fatalln(err)
+	}
+	result.Body.Close()
+	fmt.Println(string(content))
+	return (string(content))
+}
```
jgengo, line 22: This is pretty nice, but I would recommend avoid creating a function like this. It doesn't help to maintain your code if for example you are asked to also use a response header or response status code, you will have to tricks to use it.

chall03/hopham.go
```diff
@@ -0,0 +1,48 @@
...
+
+func respondToURL(content string) {
+	var id, r, g, b int
+
+	getInfo, err := fmt.Sscanf(string(content), "id=%d,r=%d,g=%d,b=%d - Send your response here: chall03.hive.fi/?id=<id>&resp=<hex>", &id, &r, &g, &b)
```
jgengo, line 27: nice you found Sscanf

jgengo (approved):

PR #140, opened 2020-04-02
chall03 submit

Rouhija:
Everywhere we use errCheck utility to handle errors. In case of errors we use Fatalf to print out the error and exit afterwards. Also both resp bodies are deferred to close when the program exits.

1. Get request to the initial URL
2. Read bytes from http response body and send it to parseResp()
3. Match the wanted values with regex FindAllStringSubmatch (id, r, g, b)
4. Convert rgb values with Atoi to int
5. Use Sprintf to format the second get request (id as is, r, g and b values to hex string with 2 width)
6. Log the status code from the second request

### DIFF
--- a/chall03/srouhe.go
+++ b/chall03/srouhe.go
@@ -1,0 +1,45 @@
+package main
+
+import (
+	"fmt"
+	"log"
+	"regexp"
+	"strconv"
+	"net/http"
+	"io/ioutil"
+)
+
+
+func errCheck(err error) {
+	if err != nil {
+        log.Fatalf("Fatal error: %v", err)
+    }
+}
+
+func parseResp(respBody string) (string) {
+
+	re, err := regexp.Compile("id=(\\d*).*r=(\\d{1,3}).*g=(\\d{1,3}).*b=(\\d{1,3})")
+	errCheck(err)
+	matches := re.FindAllStringSubmatch(respBody, -1)
+	r, err := strconv.Atoi(matches[0][2])
+	errCheck(err)
+	g, err := strconv.Atoi(matches[0][3])
+	errCheck(err)
+	b, err := strconv.Atoi(matches[0][4])
+	errCheck(err)
+	return (fmt.Sprintf("https://chall03.hive.fi/?id=%v&resp=%.2x%.2x%.2x", matches[0][1], r, g, b))
+}
+
+func main() {
+	resp, err := http.Get("https://chall03.hive.fi/")
+	defer resp.Body.Close()
+	errCheck(err)
+
+	body, err := ioutil.ReadAll(resp.Body)
+	errCheck(err)
+
+	response, err := http.Get(parseResp(string(body)))
+	defer response.Body.Close()
+	errCheck(err)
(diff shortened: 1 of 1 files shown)


chall03/srouhe.go
```diff
@@ -0,0 +1,45 @@
...
+	"net/http"
+	"io/ioutil"
+)
+
+
+func errCheck(err error) {
+	if err != nil {
+        log.Fatalf("Fatal error: %v", err)
+    }
+}
```
jgengo, line 17: The idea is good, but unfortunately, this is not following the Golang best practice. It can sound weird because it forces you to go against the DRY concept, but it's better to keep your error handling right under where you use it.

Some errors could just raise a warning and not os.Exit(1) (what log.Fatalf is doing) so this function wouldn't work.

chall03/srouhe.go
```diff
@@ -0,0 +1,45 @@
...
+
+func parseResp(respBody string) (string) {
+
+	re, err := regexp.Compile("id=(\\d*).*r=(\\d{1,3}).*g=(\\d{1,3}).*b=(\\d{1,3})")
```
jgengo, line 21: so far, the best regex I've seen; but if you plan to do something like that you should maybe use `fmt.Sscanf()` instead of regex.

I suggest reading: https://softwareengineering.stackexchange.com/a/113243

chall03/srouhe.go
```diff
@@ -0,0 +1,45 @@
...
+	g, err := strconv.Atoi(matches[0][3])
+	errCheck(err)
+	b, err := strconv.Atoi(matches[0][4])
+	errCheck(err)
+	return (fmt.Sprintf("https://chall03.hive.fi/?id=%v&resp=%.2x%.2x%.2x", matches[0][1], r, g, b))
+}
+
+func main() {
+	resp, err := http.Get("https://chall03.hive.fi/")
+	defer resp.Body.Close()
```
jgengo, line 35: defer should be after the error check. 

```
General concept is that when a function (or method) has multi return values one being an error, error should be checked first and only proceed if the error is nil. Functions should return zero values for other (non-error) values if there is an error. If the function behaves differently, it should be documented. http.Get() does not document such deviation.
```

chall03/srouhe.go
```diff
@@ -0,0 +1,45 @@
...
+	errCheck(err)
+
+	body, err := ioutil.ReadAll(resp.Body)
+	errCheck(err)
+
+	response, err := http.Get(parseResp(string(body)))
+	defer response.Body.Close()
```
jgengo, line 42: Same here

```
General concept is that when a function (or method) has multi return values one being an error, error should be checked first and only proceed if the error is nil. Functions should return zero values for other (non-error) values if there is an error. If the function behaves differently, it should be documented. http.Get() does not document such deviation.
```

jgengo (commented):
Good job!

chall03/srouhe.go
Rouhija: Thanks for helpful feedback! Sscanf() would've been smoother indeed and good points in the article. Hopefully there's plenty of go in upcoming challenges :)

PR #141, opened 2020-04-02
Add golang challenge

hakolao:
1: Suck at golang, try using c functions from go
=> Too slow, cry
2: Retry


jgengo:
Unfortunately, your PR description isn't valid enough.

PR #142, opened 2020-04-02
chall03

lkallio:
The most interesting thing about golang based on this chall: multiple return values for functions. I like that.


chall03/lkallio.go
```diff
@@ -0,0 +1,52 @@
...
+		log.Fatal(err)
+	}
+	out, err := ioutil.ReadAll(res.Body)
+	res.Body.Close()
+	if err != nil {
+		log.Fatal(err)
+	}
+	return string(out)
+}
+
```
jgengo, line 22: It's nice.

chall03/lkallio.go
```diff
@@ -0,0 +1,52 @@
...
+		loop_phase += 1
+	}
+	return id, hex
+}
+
+func main() {
+	id, hex := get_values(retrieve_resp("https://chall03.hive.fi"))
+	id_string := fmt.Sprintf("%d", id)
+	hex_string := fmt.Sprintf("%x", hex)
+	fmt.Println(retrieve_resp("https://chall03.hive.fi/?id=" + id_string + "&resp=" + hex_string))
```
jgengo, line 51: I suggest using gofmt

The problem with your function retrieve_resp, it's not really handy if you want to get some other information from the response. (like that status_code, etc). Won't help to maintain your code.

jgengo (approved):

PR #144, opened 2020-04-02
rjaakonm chall03

rasmusjaa:
First time doing Golang, very nice challenge!

1. Start timer
2. GET url, and get content to string and print it
3. Split string with " " to get value part and response url
4. Split value part with "=" or "," to get values
5. Convert RGB values to hex. Printf after atoi would have been easier, but I wanted to try something new with Go, so I'm using FormatInt and padding manually if needed
6. Replace id and hex from response url
7. GET new url, and get content to string and print it
8. Print time since start

### DIFF
--- a/chall03/rjaakonm.go
+++ b/chall03/rjaakonm.go
@@ -5,7 +5,6 @@ import (
 	"io/ioutil"
 	"log"
 	"net/http"
-	"os"
 	"strconv"
 	"strings"
 	"time"
@@ -18,7 +17,6 @@ func mySplit(r rune) bool {
 func myErrorCheck(err error) {
 	if err != nil {
 		log.Fatal(err)
-		os.Exit(1)
 	}
 }
 
--- a/chall03/rjaakonm.go
+++ b/chall03/rjaakonm.go
@@ -54,13 +54,9 @@ func main() {
 	content := myGetResponse(url)
 	fmt.Println("answer: " + content)
 
-	// Split string with " " to get value part and response url
 	parts := strings.Split(string(content), " ")
-	// Split value part with "=" or "," to get values
 	subparts := strings.FieldsFunc(parts[0], mySplit)
-	// Convert RGB values to hex, trying without printf
 	strhex := myColorConverter(subparts[3], subparts[5], subparts[7])
-	// Replace id and hex from response url
 	r := strings.NewReplacer("<id>", subparts[1], "<hex>", strhex)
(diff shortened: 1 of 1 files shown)


chall03/rjaakonm.go
```diff
@@ -0,0 +1,65 @@
...
+
+func mySplit(r rune) bool {
+	return r == '=' || r == ','
+}
+
+func myErrorCheck(err error) {
+	if err != nil {
+		log.Fatal(err)
+	}
+}
```
jgengo, line 21: You should not do that in Go.

I know it seems to go against the DRY concept, but error handling you always be right after the usage. This function is going against the best practice of Go.

chall03/rjaakonm.go
```diff
@@ -0,0 +1,65 @@
...
+
+func main() {
+	started := time.Now()
+
+	url := "https://chall03.hive.fi/"
+	content := myGetResponse(url)
+	fmt.Println("answer: " + content)
+
+	parts := strings.Split(string(content), " ")
+	subparts := strings.FieldsFunc(parts[0], mySplit)
```
jgengo, line 56: nice usage.

mySplit could have been an anonymous function / close function inside main()

chall03/rjaakonm.go
```diff
@@ -0,0 +1,65 @@
...
+	if err != nil {
+		log.Fatal(err)
+	}
+}
+
+func myGetResponse(str string) string {
+	resp, err := http.Get(str)
+	myErrorCheck(err)
+	content, err := ioutil.ReadAll(resp.Body)
+	defer resp.Body.Close()
```
jgengo, line 27: I think `defer` isn't necessary here.

chall03/rjaakonm.go
```diff
@@ -0,0 +1,65 @@
...
+	parts := strings.Split(string(content), " ")
+	subparts := strings.FieldsFunc(parts[0], mySplit)
+	strhex := myColorConverter(subparts[3], subparts[5], subparts[7])
+	r := strings.NewReplacer("<id>", subparts[1], "<hex>", strhex)
+	newurl := r.Replace(parts[6])
+
+	content = myGetResponse("https://" + newurl)
+	fmt.Println("answer: " + content)
+
+	fmt.Printf("DONE: %dms\n", time.Since(started).Milliseconds())
```
jgengo, line 64: I like the fact you have a diff time.

jgengo (approved):

PR #145, opened 2020-04-02
go

Rakutaa:
if error happens in req or read, log and exit

if everything goes smoothly, 
getting response and making it to a string and parse ints from there and changing RGB to hex and making new req.

### DIFF
--- a/chall03/vtran.go
+++ b/chall03/vtran.go
@@ -1,0 +1,38 @@
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
+	"log"
+	"net/http"
+)
+
+func main() {
+	resp, err := http.Get("https://chall03.hive.fi/")
+	ifError(err)
+	defer resp.Body.Close()
+
+	if resp.StatusCode == http.StatusOK {
+		bodyBytes, err := ioutil.ReadAll(resp.Body)
+		ifError(err)
+
+		s := parseStringToQuerry(string(bodyBytes))
+		res, err := http.Get(s)
+		ifError(err)
+
+		defer res.Body.Close()
+		println(res.StatusCode)
+	}
+}
+
+func parseStringToQuerry(str string) string {
+	var id, r, g, b int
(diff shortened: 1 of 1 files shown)


chall03/vtran.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"fmt"
+	"io/ioutil"
+	"log"
+	"net/http"
+)
+
+func main() {
+	resp, err := http.Get("https://chall03.hive.fi/")
+	ifError(err)
```
jgengo, line 12: You shouldn't do that in Go, I know that if err != nil {} can be a bit repetitive, and makes the code longer. It doesn't respect the best usage of Go.

chall03/vtran.go
```diff
@@ -0,0 +1,38 @@
...
+		ifError(err)
+
+		defer res.Body.Close()
+		println(res.StatusCode)
+	}
+}
+
+func parseStringToQuerry(str string) string {
+	var id, r, g, b int
+	fmt.Sscanf(str, "id=%d,r=%d,g=%d,b=%d", &id, &r, &g, &b)
```
jgengo, line 30: smart move!

chall03/vtran.go
```diff
@@ -0,0 +1,38 @@
...
+	return fmt.Sprintf("https://chall03.hive.fi/?id=%d&resp=%02x%02x%02x", id, r, g, b)
```
jgengo, line 31: good!

chall03/vtran.go
```diff
@@ -0,0 +1,38 @@
...
+
+	if resp.StatusCode == http.StatusOK {
+		bodyBytes, err := ioutil.ReadAll(resp.Body)
+		ifError(err)
+
+		s := parseStringToQuerry(string(bodyBytes))
+		res, err := http.Get(s)
+		ifError(err)
+
+		defer res.Body.Close()
```
jgengo, line 23: you don't need to use defer here while it's supposed to be close here.

jgengo (approved):
Good job.

PR #147, opened 2020-04-03
chall03 - mlindhol

MikeyLHolm:
After getting first response I split the string from "," and then with regexp remove all non-numeric chars. Last str[3] has no "," so have to trim the 03 from chall03 at the end of the string. Atoi them to ints and then make another hex with sprintf. 

Then another request to get the response and a StatusCode.

### DIFF
--- a/chall03/mlindhol.go
+++ b/chall03/mlindhol.go
@@ -1,0 +1,55 @@
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"regexp"
+	"strconv"
+	"strings"
+)
+
+func error() {
+	fmt.Printf("whoops, error!\n")
+	os.Exit(1)
+}
+
+func main() {
+
+	resp, err := http.Get("https://chall03.hive.fi/")
+	if err != nil {
+		error()
+	}
+
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		error()
+	}
+
+	str := strings.Split(string(body), ",")
+
+	reg, err := regexp.Compile("[^0-9]+")
+	if err != nil {
+		error()
+	}
+
+	id, err := strconv.Atoi(reg.ReplaceAllString(str[0], ""))
+	r, err := strconv.Atoi(reg.ReplaceAllString(str[1], ""))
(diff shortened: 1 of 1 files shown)


chall03/mlindhol.go
```diff
@@ -0,0 +1,55 @@
...
+	"os"
+	"regexp"
+	"strconv"
+	"strings"
+)
+
+func error() {
+	fmt.Printf("whoops, error!\n")
+	os.Exit(1)
+}
```
jgengo, line 16: for the line 14, you should use fmt.Println() if you don't have variable and want to print a new line.

Plus, this function already exist in Go:

```go
log.Fatalln("whoops, error!")
```

chall03/mlindhol.go
```diff
@@ -0,0 +1,55 @@
...
+	resp, err := http.Get("https://chall03.hive.fi/")
+	if err != nil {
+		error()
+	}
+
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		error()
+	}
+
```
jgengo, line 29: resp.Body should be closed after usage.

resp.Body.Close() here.

chall03/mlindhol.go
```diff
@@ -0,0 +1,55 @@
...
+	str := strings.Split(string(body), ",")
+
+	reg, err := regexp.Compile("[^0-9]+")
```
jgengo, line 32: You should avoid using regex: https://softwareengineering.stackexchange.com/a/113243

chall03/mlindhol.go
```diff
@@ -0,0 +1,55 @@
...
+	if err != nil {
+		error()
+	}
+
+	hex := fmt.Sprintf("%02x%02x%02x", r, g, b)
+	resp, err = http.Get(fmt.Sprintf("https://chall03.hive.fi/?id=%d&resp=%s", id, hex))
+	if err != nil {
+		error()
+	}
+	response, err := ioutil.ReadAll(resp.Body)
```
jgengo, line 50: resp.Body should be closed here as well.

jgengo (approved):
Good job.

It would have been nice to see a diff time step by step. It's not really hard to implement in Go.

```go
now := time.Now()
end := time.Since(now).Milliseconds 
```

PR #151, opened 2020-04-06
sadawi.py

salahadawi:
I save the input into a 2d list so i can directly modify each character. In a for loop, I check each character of the saved input, and when a sand block is found, i move it down one step if possible. Once a loop has been gone through without any blocks moving, the simulation is done and the end result is printed.

### DIFF
--- a/chall04/sadawi.py
+++ b/chall04/sadawi.py
@@ -1,6 +1,16 @@
 #! /usr/bin/python3
 
-i = int(input())
+import sys
+
+def error(str):
+	print("Error: " + str)
+	sys.exit()
+
+try:
+	i = int(input())
+except:
+	error("Input is not a digit.")
+	sys.exit()
 
 arr = []
 for j in range(i):
--- a/chall04/sadawi.py
+++ b/chall04/sadawi.py
@@ -1,4 +1,5 @@
 #! /usr/bin/python3
+#Python 3.6.9
 
 import sys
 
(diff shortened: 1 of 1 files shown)


chall04/sadawi.py
```diff
@@ -0,0 +1,34 @@
...
+
+try:
+	i = int(input())
+except:
+	error("Input is not a digit.")
+	sys.exit()
+
+arr = []
+for j in range(i):
+	arr.append(list(input()))
```
jgengo, line 18: Can be written the pythonic way: 

```python
arr = [list(input()) for _ in range(i)]
```

Plus, you are not using `j` so you should move it to a blank identifier. `_`

chall04/sadawi.py
```diff
@@ -0,0 +1,34 @@
+#! /usr/bin/python3
```
jgengo, line 1:
```
#!/usr/bin/env python3
```

chall04/sadawi.py
```diff
@@ -0,0 +1,34 @@
...
+
+def error(str):
+	print("Error: " + str)
+	sys.exit()
+
+try:
+	i = int(input())
+except:
+	error("Input is not a digit.")
+	sys.exit()
```
jgengo, line 14: error(), already sys.exit, so the following sys.exit() line 14 is not necessary.

error() is only used once, so I would avoid creating a function for that and you could use sys.exit("message") ; https://docs.python.org/2/library/sys.html#sys.exit

jgengo (approved):
Good job,

I would really appreciate you send your PR once you sure your code is done. 

It's also good for you to take advantage of those exercises to learn to propose pull-request the right way for the future.

PR #152, opened 2020-04-06
Falling sand

kimsappi:
First I read and validate the input (all rows must have N characters and there must only be valid characters).

Then I take a checksum of the array I store the input into. Then I enter an infinite loop where I go from the bottom to the top and check if there is a sand particle that can fall down (by checking the lower row). If it can fall, I slice the strings so that I move the sand down. After every full iteration I check if the checksum matches the old one (I should have used a bool flag here, but whatever). When they match, the array is unchanged, and the simulation is complete. Then I print the output.

### DIFF
--- a/chall04/ksappi.py
+++ b/chall04/ksappi.py
@@ -1,0 +1,53 @@
+#!/usr/bin/python3
+
+# Version: python3.7
+
+import sys
+import re
+import hashlib
+
+def calculateHashOfArray(array):
+	arrayStr = ''
+	for row in array:
+		arrayStr += row
+	return hashlib.md5(arrayStr.encode('utf-8')).hexdigest()
+
+try:
+	size = int(input())
+except:
+	print('Give the size of the board as an integer')
+	sys.exit()
+board = []
+if size == 0:
+	sys.exit()
+for i in range(size):
+	board.append(input())
+
+for row in board:
+	if not re.match('^[ \.#]+$', row):
+		print('Enter only valid characters (" ", ".", "#")')
+		sys.exit()
+	if len(row) != size:
+		print('Enter a complete row')
+		sys.exit()
+
+oldChecksum = calculateHashOfArray(board)
+while True:
+	for rowIndex in range(len(board) - 2, -1, -1):
+		row = board[rowIndex]
+		print (row, '###')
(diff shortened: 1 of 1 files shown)


chall04/ksappi.py
```diff
@@ -0,0 +1,51 @@
...
+	arrayStr = ''
+	for row in array:
+		arrayStr += row
+	return hashlib.md5(arrayStr.encode('utf-8')).hexdigest()
+
+try:
+	size = int(input())
+except:
+	print('Give the size of the board as an integer')
+	sys.exit()
```
jgengo, line 19: you can use sys.exit("message") ; https://docs.python.org/2/library/sys.html#sys.exit

chall04/ksappi.py
```diff
@@ -0,0 +1,51 @@
...
+board = []
+if size == 0:
+	sys.exit()
+for i in range(size):
+	board.append(input())
```
jgengo, line 24: The pythonic way to write
```python
board = []
for i in range(size):
	board.append(input())
```
is
```python
board = [list(input()) for _ in range(size)]
```

Plus, you are not using `i` so I would recommend to move it a to a blank identifier.

chall04/ksappi.py
```diff
@@ -0,0 +1,51 @@
...
+
+for row in board:
+	if not re.match('^[ \.#]+$', row):
+		print('Enter only valid characters (" ", ".", "#")')
```
jgengo, line 28: I tend to suggest avoid regex as much as possible. However, I never really benchmarked it.

chall04/ksappi.py
```diff
@@ -0,0 +1,51 @@
...
+					nextRow = nextRow[:i] + '.' + nextRow[i + 1:]
+					board[rowIndex + 1] = nextRow
+					board[rowIndex] = board[rowIndex][:i] + ' ' + board[rowIndex][i + 1:]
+	checksum = calculateHashOfArray(board)
+	if checksum == oldChecksum:
+		break
+	oldChecksum = checksum
+
+for row in board:
+	print(row)
```
jgengo, line 51: this could also be written to the pythonic way.

jgengo (approved):
Good job!

PR #153, opened 2020-04-06
Add my solution to chall04

hiiri:
Store input into 2D list, each list containing the characters of each input line. Check at every sand character if the spot under it (next row, same column) is a space. If yes, set that position to be sand and change current position to a space.
Continue this until no sand has moved and we know every sand has fallen as far as it can.

### DIFF
--- a/chall04/alcohen.py
+++ b/chall04/alcohen.py
@@ -1,5 +1,7 @@
 #!/usr/bin/python3
 
+# Version: Python 3.6.9
+
 size = int(input())
 grid = []
 
--- a/chall04/alcohen.py
+++ b/chall04/alcohen.py
@@ -11,15 +11,14 @@ def can_move(next_under):
         return True
     return False
 
-moved = True
-while moved:
-    moved = False
-    for row in range(size):
-        for col in range(size):
-            if (grid[row][col] == '.'):
-                if (row + 1 < size and can_move(grid[row + 1][col])):
-                    grid[row][col] = ' '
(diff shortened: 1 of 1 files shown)


chall04/alcohen.py
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/python3
+
+# Version: Python 3.6.9
+
+def error_input(msg):
+    print(msg)
+    quit()
```
jgengo, line 7: instead of creating your own function `error_input` you could use `sys.exit("your message")`

chall04/alcohen.py
```diff
@@ -0,0 +1,42 @@
...
+
+def can_move(next_under):
+    if (next_under == ' '):
+        return True
+    return False
+
+if __name__ == '__main__':
```
jgengo, line 14: I would recommend putting the whole block inside a function main. But it's really good you protected it.

chall04/alcohen.py
```diff
@@ -0,0 +1,42 @@
...
+    size = input()
+    if size.isdigit():
+        size = int(size)
+    else:
+        error_input("Enter a positive number")
+
+    grid = [list(input()) for i in range(size)] # Split 'size' amount of lines into lists of characters
```
jgengo, line 21: nice!

However, i is not used so you should move it to a blank identifier `_`

```python
grid = [list(input()) for _ in range(size)]
```

jgengo (approved):
Good job!

PR #154, opened 2020-04-06
Chall04 solution

tdawson-42:
I first read the input size and check it is given as a non-negative integer. Then I read lines according to the size and check that each input row is the correct length and contains only the allowed characters. I create a 2d array from the input and then iterate through it starting from the second last row and moving up. Every time I come across a sand character, I check if the character below it is a space. If so, I swap the characters with one another so the sand moves down one position.  This happens in a while loop with a boolean value used to check if any sand was moved. If not, the loop ends.

### DIFF
--- a/chall04/tdawson.py
+++ b/chall04/tdawson.py
@@ -1,23 +1,33 @@
 #!/usr/bin/python3
-#Python 3.6.9
+# Python 3.6.9
 
 import sys
 
-n = int(input())
-grid = []
+def error(error_msg):
+    sys.exit('Error: ' + error_msg)
 
-for _ in range (n):
-    grid.append(list(input()))
+size = input()
+if not (size.isdigit()):
+    error('Not a positive integer.')
+size = int(size)
+
+grid = []
+for _ in range(size):
+    row = input()
+    if len(row) != size:
+        error('Input length different than size.')
+    if not all(c in ' #.' for c in row):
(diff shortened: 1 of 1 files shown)


chall04/tdawson.py
```diff
@@ -0,0 +1,28 @@
...
+size = input()
+if not (size.isdigit()):
+    sys.exit('Error: Not a positive integer.')
+size = int(size)
+
+grid = []
+for _ in range(size):
+    row = input()
+    if len(row) != size or not all(c in ' #.' for c in row):
+        sys.exit('Error: Inputted row is not valid.')
```
jgengo, line 15: nice usage of sys.exit argument.

chall04/tdawson.py
```diff
@@ -0,0 +1,28 @@
...
+while not (settled):
+    settled = True
+    for row in range(size - 2, -1, -1):
+        for col, c in enumerate(grid[row]):
+            if (c is '.' and grid[row+1][col] is ' '):
+                grid[row][col] = ' '
+                grid[row+1][col] = '.'
+                settled = False
+
+print('\n'.join(''.join(row) for row in grid))
```
jgengo, line 28: clean!

jgengo (approved):
Clean code!

tiny suggestion, I would just recommend separating the grid appending and the check.

PR #155, opened 2020-04-06
Chall04 (jnovotny)

jiricodes:
Firstly I check whether given N is an integer, then if the amount of read lines is  N.

I've created a class to have easier time testing and doing small changes. The class transforms raw input into a 2 dimensional array self.map and checks for its validity (only allowed symbols and each line has N length).

Then the actual simulation is done in function simulate() where I traverse the self.map in reverse and if current tile is free and tile 'above' contains 'sand' the values are adjusted. This is repeated until there are no changes done, at which point the function exits.

Then the map is printed using class function __str__().

### DIFF
--- a/chall04/jnovotny.py
+++ b/chall04/jnovotny.py
@@ -1,0 +1,61 @@
+import sys
+
+def map_error_exit(msg):
+	print(f"Error occured: {msg}")
+	exit()
+
+class FallingSand():
+	def __init__(self, size, raw):
+		self.size = size
+		self.map = list()
+		self.setup(raw)
+	
+	def __str__(self):
+		res = ""
+		for row in self.map:
+			for c in row:
+				res += c
+			res += '\n'
+		return res.rstrip('\n')
+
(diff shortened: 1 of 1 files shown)


chall04/jnovotny.py
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/python3
+#Python 3.7.3
+
+import sys
+
+def map_error_exit(msg):
+	print(f"Error occured: {msg}")
+	exit()
```
jgengo, line 8: I really love the new format of python3!

I would suggest using sys.exit(arg) instead of creating your own function. 

chall04/jnovotny.py
```diff
@@ -0,0 +1,64 @@
...
+
+class FallingSand():
```
jgengo, line 10: wuw! a class! really nice

jgengo (approved):
That's really clean.

You did a great job here with the class, I appreciate you dig deeper in Python to make this exercise.

PR #156, opened 2020-04-06
falling sand - rjaakonm

rasmusjaa:
Tried to make it with minimum loops through content, so I have 3 loops: loop input to list, loop through each world line once to drop sand, loop through each world line to print lines. Couldn't figure out a way to combine any of them, since on first loop I have to have next line available too and on second loop I don't want to print anything since there might still be errors later. One possible optimization might have been to join lines with '\n' and call just one print, but not sure if join x amount of lines or print x times is more effective.

1. Read lines from standard input and append them to list as character lists
2. Convert first line to int and check that it matches simulated world line count
3. Loop through all list lines and in them all list chars. If char is sand and on same position in next line there is empty, drop sand (switch sand to empty and empty to sand).
4. Join each character list to string after it has been iterated through.
5. Print strings from the list (all but the first that has number)

Error handling: Return if first line can't be converted to int. Return if world line count is different than the number on first. Line must have number amount of characters '.', ' ' or '#' and end in line change or string terminator, otherwise return.

### DIFF
--- a/chall04/rjaakonm.py
+++ b/chall04/rjaakonm.py
@@ -8,8 +8,8 @@ def main():
 	for line in sys.stdin:
 		lines.append(list(line))
 		counted += 1
-	lines[0] ="".join(lines[0])
 	try:
+		lines[0] ="".join(lines[0])
 		size = int(lines[0])
 	except:
 		return
--- a/chall04/rjaakonm.py
+++ b/chall04/rjaakonm.py
@@ -7,7 +7,10 @@ def main():
 	for line in sys.stdin:
 		lines.append(list(line))
 	lines[0] ="".join(lines[0])
-	size = int(lines[0])
+	try:
+		size = int(lines[0])
+	except:
+		return
 	if size == len(lines) - 1:
 		i = 1
 		while i < size:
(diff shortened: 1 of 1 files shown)


chall04/rjaakonm.py
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/python3
+
+import sys
+
+def main():
+	lines = []
+	counted = 0
+	for line in sys.stdin:
+		lines.append(list(line))
+		counted += 1
```
jgengo, line 10: This could be written in a more pythonic way.

```python
grid = [list(input()) for _ in range(n)]
```
and I think `counted` could be use another way: `len(grid)`

chall04/rjaakonm.py
```diff
@@ -0,0 +1,41 @@
...
+			lines[i] = "".join(lines[i])
+			i += 1
+		lines[size] = "".join(lines[size])
+		i = 1
+		while i <= size:
+			print(lines[i])
+			i += 1
+
+if __name__ == "__main__":
+    main()
```
jgengo, line 41: good that you protected it.

jgengo (approved):
Great job, your PR description is good as well!

PR #157, opened 2020-04-06
Submission

Rouhija:
Made it with class just for practice

First the class is initialized and size and input are parsed to array of arrays, which checks for correct characters and line lengths. Lines after size are ignored.

The if there is a valid size and map is not empty, we call drop_sand() which goes through, x coord at the time, grains of sand and drops them line by line, breaking the loop when it's stuck.

Then we call print_world to display the resulting map.

### DIFF
--- a/chall04/srouhe.py
+++ b/chall04/srouhe.py
@@ -1,0 +1,48 @@
+#!/usr/bin/python3
+
+import sys
+
+def split(word): 
+    return [char for char in word]
+
+def parse_stdin():
+	world = []
+	count = 0
+	try:
+		for line in sys.stdin:
+			if count == 0:
+				size = int(line)
+			elif count <= size:
+				world.append(split(line.replace('\n', '')))
+			count += 1
+	except Exception as e:
+		print(f'Error parsing input: [{e}]')
+		exit(1)
+	return size, world
+
+def check_chars(c1, c2):
+	if c1 not in '. #' or c2 not in '. #':
+		print('Invalid character')
+		exit(1)
+
+def sand_dealer(size: int, world: list):
+	for y in range(0, size - 1):
+		for x in range(0, size):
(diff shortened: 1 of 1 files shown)


chall04/srouhe.py
```diff
@@ -0,0 +1,75 @@
...
+	def parse_world(self):
+		world = []
+		try:
+			for i, line in enumerate(sys.stdin):
+				world.append(self.check_chars(line, self.size))
+				if i > self.size - 2:
+					break
+		except Exception as e:
+			print(f'Error parsing input: [{e}]')
+			exit(1)
```
jgengo, line 38: you can use sys.exit(arg) ; https://docs.python.org/2/library/sys.html#sys.exit

chall04/srouhe.py
```diff
@@ -0,0 +1,75 @@
...
+		return line
+
+	# Preprocessing
+	def parse_size(self):
+		try:
+			for line in sys.stdin:
+				return int(line)
+		except Exception as e:
+			print(e)
+			exit(1)
```
jgengo, line 60: sys.exit(arg)

chall04/srouhe.py
```diff
@@ -0,0 +1,75 @@
...
+
+	# Utility
+	def split(self, word): 
+		return [char for char in word]
+
+
+
+
+if __name__ == '__main__':
```
jgengo, line 69: good protection

chall04/srouhe.py
```diff
@@ -0,0 +1,75 @@
...
+
+	world = SandDealer()
+	if world.size is not None and world.world != []:
```
jgengo, line 72: not sure here, if it's to check world is an empty list you can also do 

`and world.world`

jgengo (approved):
I really appreciate you did take the time to explore this exercise with class.

PR #158, opened 2020-04-06
chall04

vkuikka:
First takes size of the world array as input. Then appends all correct length inputs to world array until that array has the right amount of right length strings. Iterates through those strings from bottom to top while moving all found dots down. Then prints the result array at the end.

### DIFF
--- a/chall04/vkuikka.py
+++ b/chall04/vkuikka.py
@@ -1,0 +1,27 @@
+#!/usr/bin/python3
+import sys
+
+def main():
+	size = int(input())
+	world = []
+
+	if size == 0:
+		sys.exit()
+	for i in range(size):
+		world.append(input())
+		if len(world[i]) != size:
+			sys.exit()
+	for y in reversed(range(size)):
+		for x, c in enumerate(world[y]):
+			if c == '.':
+				tmp_height = y
+				world[tmp_height] = world[tmp_height][:x] + " " + world[tmp_height][x + 1:]
+				while tmp_height < size and world[tmp_height][x] == ' ':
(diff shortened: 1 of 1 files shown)


chall04/vkuikka.py
```diff
@@ -0,0 +1,28 @@
...
+	size = int(input())
+	world = []
+
+	if size == 0:
+		sys.exit()
+	for i in range(size):
+		world.append(input())
+		if len(world[i]) != size:
+			print("Wrong length")
+			sys.exit()
```
jgengo, line 14: you can pass the print into sys.exit argument ; https://docs.python.org/2/library/sys.html#sys.exit

chall04/vkuikka.py
```diff
@@ -0,0 +1,28 @@
...
+	for y in reversed(range(size)):
```
jgengo, line 15: super smart!

chall04/vkuikka.py
```diff
@@ -0,0 +1,28 @@
...
+				tmp_height = y
+				world[tmp_height] = world[tmp_height][:x] + " " + world[tmp_height][x + 1:]
+				while tmp_height < size and world[tmp_height][x] == ' ':
+					tmp_height += 1
+				tmp_height -= 1
+				world[tmp_height] = world[tmp_height][:x] + "." + world[tmp_height][x + 1:]
+	for line in world:
+		print(line)
+
+if __name__ == "__main__":
```
jgengo, line 27: good that you protected it.

jgengo (approved):
clean code

PR #159, opened 2020-04-06
chall04

osalmine:
First I get the input from standard input, until control-d is pressed. Then I parse the newlines out and check for bad input. I put spaces at the end of each line if needed so that there is as many characters in each line as the input defined is the width to be. Then I reverse the list so I can pull the sand from the bottom-most row to the bottom first. After checking each character in every line and pulling the sand as far down as possible, I reverse the list again and print it out.

### DIFF
--- a/chall04/osalmine.py
+++ b/chall04/osalmine.py
@@ -1,0 +1,59 @@
+#!/usr/bin/python3
+
+# Python 3.8.2
+
+import sys
+import fileinput
+
+def append_space(list, size):
+    for i, s in enumerate(list):
+        if len(s) == 0:
+            list[i] = "     "
+        elif len(s) < size:
+            amount = size - len(s)
+            for _ in range(amount):
+                s += ' '
+            list[i] = s
+        elif len(s) > size:
+            raise Exception("line '{0}' shouldn't have more then {1} characters".format(s, size))
+    return list
+
+def replace_sand(list, i, j, size):
+    ind = len(list) - i - 1
+    for k, search in enumerate(list[ind:]):
+        if ind + k is size or list[ind + k + 1][j] is not ' ':
+            tmp = list[ind][:j] + ' ' + list[ind][j + 1:]
+            list[ind] = tmp
+            tmp = list[ind + k][:j] + '.' + list[ind + k][j + 1:]
+            list[ind + k] = tmp
+            return list
+
+def main():
+    inp = []
+    for line in fileinput.input():
+        inp.append(line)
+    list = [x[:-1] for x in inp]
+    for str in list[1:]:
+        for char in str:
+            if char != ' ' and char != '#' and char != '.':
+                raise Exception("input can only contain ' ', '#' and '.'")
+    if not list[0].isdigit():
+        raise Exception("size can't be other than digits")
+    size = int(list[0])
+    append_space(list, size)
+    list.reverse()
+    for i, line in enumerate(list):
+        if i is size or i is 0:
+            continue
+        for j, char in enumerate(line):
+            if char is '.':
+                list.reverse()
+                list = replace_sand(list, i, j, size)
+                list.reverse()
+    list.reverse()
+    print("\n")
+    for line in list[1:]:
+        print(line)
+
+if __name__ == '__main__':
+    main()


chall04/osalmine.py
```diff
@@ -0,0 +1,59 @@
...
+                list.reverse()
+                list = replace_sand(list, i, j, size)
+                list.reverse()
+    list.reverse()
+    print("\n")
+    for line in list[1:]:
+        print(line)
+
+if __name__ == '__main__':
+    main()
```
d15ky, line 59: good job for protection from a run on import!

chall04/osalmine.py
```diff
@@ -0,0 +1,59 @@
...
+            tmp = list[ind][:j] + ' ' + list[ind][j + 1:]
+            list[ind] = tmp
+            tmp = list[ind + k][:j] + '.' + list[ind + k][j + 1:]
+            list[ind + k] = tmp
+            return list
+
+def main():
+    inp = []
+    for line in fileinput.input():
+        inp.append(line)
```
d15ky, line 34: can be achieved in a more concise way:
```
inp = sys.stdin.readlines()
```

chall04/osalmine.py
```diff
@@ -0,0 +1,59 @@
...
+    list = [x[:-1] for x in inp]
```
d15ky, line 35: if you want to strip '\n' it's better to just strip '\n'
```
list = [x.strip('\n') for x in inp]
```
or
```
list = list(map(lambda x: x.strip('\n'), inp))
```

chall04/osalmine.py
```diff
@@ -0,0 +1,59 @@
...
+    for str in list[1:]:
+        for char in str:
+            if char != ' ' and char != '#' and char != '.':
```
d15ky, line 38: faster way to achieve it would be:
```
if set(str) > set('#. '): 
```
https://www.geeksforgeeks.org/sets-in-python/

chall04/osalmine.py
```diff
@@ -0,0 +1,59 @@
...
+                raise Exception("input can only contain ' ', '#' and '.'")
```
d15ky, line 39: Good job on learning and using Exceptions! 
However, if it's a predictable error from user input, better to just quit a program with a message to stderr. Also, about the Exceptions — I suggest reading about different kinds of standard Exceptions (e.g., in this case, it would make more sense to use ValueError one)

chall04/osalmine.py
```diff
@@ -0,0 +1,59 @@
...
+    if not list[0].isdigit():
+        raise Exception("size can't be other than digits")
+    size = int(list[0])
```
d15ky, line 42: If you'd do that before the previous check, you could also add a size check when you iterate on actual data, rather than doing it in different function (which is not an explicit way to do that)

chall04/osalmine.py
```diff
@@ -0,0 +1,59 @@
...
+
+# Python 3.8.2
+
+import sys
+import fileinput
+
+def append_space(list, size):
+    for i, s in enumerate(list):
+        if len(s) == 0:
+            list[i] = "     "
```
d15ky, line 11: Did you mean `list[i] = " " * size` ?

chall04/osalmine.py
```diff
@@ -0,0 +1,59 @@
...
+        elif len(s) < size:
+            amount = size - len(s)
+            for _ in range(amount):
+                s += ' '
```
d15ky, line 15: `s += ' ' * amount`
or even
`list[i] += " " * (size - len(s))`

chall04/osalmine.py
```diff
@@ -0,0 +1,59 @@
...
+            if char != ' ' and char != '#' and char != '.':
+                raise Exception("input can only contain ' ', '#' and '.'")
+    if not list[0].isdigit():
+        raise Exception("size can't be other than digits")
+    size = int(list[0])
+    append_space(list, size)
+    list.reverse()
+    for i, line in enumerate(list):
+        if i is size or i is 0:
+            continue
```
d15ky, line 47: wherever you don't need to use the first element of your list, you can use `list[1:]`
or just reassigned it after you got the size `list = list[1:]`

chall04/osalmine.py
d15ky: also, don't know if your editor showed you that but `list` is a built-in function (not even a function but a data type) so using it as a variable name is not a good idea :) 
sand_list or sand_data would be more explicit

chall04/osalmine.py
```diff
@@ -0,0 +1,59 @@
...
+            for _ in range(amount):
+                s += ' '
+            list[i] = s
+        elif len(s) > size:
+            raise Exception("line '{0}' shouldn't have more then {1} characters".format(s, size))
+    return list
+
+def replace_sand(list, i, j, size):
+    ind = len(list) - i - 1
+    for k, search in enumerate(list[ind:]):
```
d15ky, line 23: I think, in this case you could use `for k in range(i)` since you don't use `search` anywhere further

chall04/osalmine.py
```diff
@@ -0,0 +1,59 @@
...
+        if ind + k is size or list[ind + k + 1][j] is not ' ':
+            tmp = list[ind][:j] + ' ' + list[ind][j + 1:]
+            list[ind] = tmp
+            tmp = list[ind + k][:j] + '.' + list[ind + k][j + 1:]
+            list[ind + k] = tmp
```
d15ky, line 28: no need to use tmp here, you can assign right away like
`list[ind] = list[ind][:j] + ' ' + list[ind][j + 1:]`

d15ky (approved):
Well done here! Hope you had fun with Python :) I suggest you discovering it further — it's even cooler and allows you to do all of that in a really simple and fast way.

PR #160, opened 2020-04-06
Chall04 falling sand

lapaset:
in my solution i
-read the input from stdin
-extract the N from input
-extract the grid from input
-as long as there is movement:
  -go through the grid starting from the bottom
  -if theres an empty spot check if the spot above is sand
     ---> if yes modify the strings so that empty spot is sand and sand is empty
     ---> if no move along
-print the result


jgengo:
Your solution was clean but unfortunately, I can't accept the PR because of the extra files.

PR #161, opened 2020-04-06
Add my solution to sand fall simulation

hakolao:
Simple int parsing validation for input, rows are not validated.

This script reads input & parses first line into a `row_size` variable & pops it out of the array. A grid string is then joined with remaining input.

`sand_fall` loops each index in the `grid`:
- if the current position has a sand piece AND next row is empty, sand piece is moved to next position
- repeat until second last row is reached
- if change took place start over
- done when no sand moves

Test by pasting this to terminal:
```
{
cat <<LOL
5
.....
  #  
#    
     
    .
LOL
} | python3 ohakola.py
```


chall04/ohakola.py
```diff
@@ -0,0 +1,50 @@
...
+#                                                                              #
+#                                                         :::      ::::::::    #
+#    ohakola.py                                         :+:      :+:    :+:    #
+#                                                     +:+ +:+         +:+      #
+#    By: ohakola <ohakola@student.hive.fi>          +#+  +:+       +#+         #
+#                                                 +#+#+#+#+#+   +#+            #
+#    Created: 2020/04/06 17:30:11 by ohakola           #+#    #+#              #
+#    Updated: 2020/04/06 17:58:25 by ohakola          ###   ########.fr        #
+#                                                                              #
+# **************************************************************************** #
```
jgengo, line 11: Please, do not add the 42 header for those challenges and try to keep it for the curriculum.

chall04/ohakola.py
```diff
@@ -0,0 +1,50 @@
...
+			print(grid[y * size + x], end = "")
+		print("\n", end = "")
+
+if __name__ == "__main__":
+	lines = sys.stdin.readlines()
+	try:
+		row_size = int(lines.pop(0))
+	except:
+		print("Wrong size given")
+		exit()
```
jgengo, line 45: your print can be passed as argument to sys.exit() ; https://docs.python.org/2/library/sys.html#sys.exit

jgengo (approved):
Sorry, but I don't have so much comment about your code. It's really clean.

PR #162, opened 2020-04-06
Minecraft physics

RandommUser:
Loop input until complete input has been gotten. First get the size by using int(), which is basically atoi. Gives error if not valid
Then start to read lines for the map. Each line gets checked for only having valid characters and the width, and it can handle multiple line-break lines (maybe not possible, but just in case) by counting all of the line-breaks.
When wanted amount of lines for the map has been read we start to move the sand.
Transform the string to a char array for easy manipulation. Start from the end. When on an air block, find sand above the spot and if found pull it down. Stops if stone is encountered. Repeate till the second from the top line has been finished reading. Then combine the array and print

Unique errors for the different bad input errors. Exception handling for the "atoi". Handles also 0 sized maps.

### DIFF
--- a/chall04/phakakos.py
+++ b/chall04/phakakos.py
@@ -1,0 +1,77 @@
+#!/usr/bin/env python3
+
+import sys
+import fileinput
+
+def err(num):
+        if (num == 0):
+                print("Give size as a number")
+                exit(1)
+        if (num == 1):
+                print("Invalid map input. Please check")
+                exit(2)
+        if (num == 2):
+                print("Map width does not match")
+                exit(3)
+
+def allowedChar(c):
+        allowed ={
+                "." : ".",
+                "#" : "#",
+                " " : " ",
+                "\n" : "\n"
+                }
+        return allowed.get(c, "error")
+
+def lineCount(line, size):
+        i = 0
+        for c in line:
+                if (allowedChar(c) == "error"):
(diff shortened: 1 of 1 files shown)


chall04/phakakos.py
```diff
@@ -0,0 +1,77 @@
...
+def err(num):
+        if (num == 0):
+                print("Give size as a number")
+                exit(1)
+        if (num == 1):
+                print("Invalid map input. Please check")
+                exit(2)
+        if (num == 2):
+                print("Map width does not match")
+                exit(3)
```
jgengo, line 15: err(1) ; err(0) ; err(2) are only used once each. So I would avoid creating a function like this one and suggest to use .

```python
sys.exit("your message")
```

chall04/phakakos.py
```diff
@@ -0,0 +1,77 @@
...
+
+def allowedChar(c):
+        allowed ={
+                "." : ".",
+                "#" : "#",
+                " " : " ",
+                "\n" : "\n"
+                }
+        return allowed.get(c, "error")
+
```
jgengo, line 25: I would not create a whole function for that. You can directly use:
```python
if not allowed.get(c):
  sys.exit("Invalid map input. Please check")
```

jgengo (approved):
Good job! your code is clean

PR #163, opened 2020-04-06
My falling sand

DianaMukaliyeva:
1. Checked with error handling if first input is integer.
2. Took next lines and checked if input is valid, then appended to an array.
3. Implemented function falling_sand, which put dots to the "empty" places

### DIFF
--- a/chall04/dmukaliy.py
+++ b/chall04/dmukaliy.py
@@ -1,0 +1,46 @@
+#!/usr/bin/python3.7
+
+import sys
+
+def falling_sand(input_array):
+    length = len(input_array)
+    y = 0
+    while y < length:
+        x = 0
+        while x < length:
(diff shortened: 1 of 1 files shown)


chall04/dmukaliy.py
```diff
@@ -0,0 +1,46 @@
...
+            return False
+    return True
+
+def main():
+    input_array = []
+    try:
+        length = int(input())
+    except:
+        print("\033[91mNot valid integer!\033[0m")
+        sys.exit(1)
```
jgengo, line 33: you can use sys.exit argument to pass the print.

```python
sys.exit("\033[91mNot valid integer!\033[0m")
```

jgengo (approved):
I'm sorry but I don't have so much to comment on your code, it's pretty clean and clear.

PR #164, opened 2020-04-06
chall04/gmolin

avocadohooman:
1. Store file / input line by line in a 2d array
2. Convert first array into a string
3. Check if N is a digit
4. Check if N is a valid number (=size of map)
5. Check each line if it is exactly the length of N
6. Check if there are no forbidden characters
7. If there is a sand, check if on the same index on the next row is either a rock or sand, if yes don't move forward, if not, then sand continues flowing
8. Concert each array into one string
9. Print line by line without new line from the print function


chall04/gmolin.py
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/python3
+
+import sys
+
+def main():
+	board = []
+	for line in sys.stdin:
+		board.append(list(line))
```
jgengo, line 8: Can be written that way:
```python
n = input()
board = [list(input()) for _ in range(n)]
```

or even 

```python
board = read.stdin.readlines()
```

chall04/gmolin.py
```diff
@@ -0,0 +1,46 @@
...
+				j += 1
+			board[i]= "".join(board[i])
+			i += 1
+		board[board_size]= "".join(board[board_size])
+		i = 1
+		while (i <= board_size):
+			print(board[i], end = "")
+			i += 1
+
+if __name__ == "__main__":
```
jgengo, line 45: good you protected it.

chall04/gmolin.py
```diff
@@ -0,0 +1,46 @@
...
+		board.append(list(line))
+	board[0] ="".join(board[0])
+	try:
+		board_size = int(board[0].rstrip())
+	except:
+		print ("Error: N is not a digit")
+		return
+	if board_size != len(board) - 1:
+		print("Error: Invalid map size")
+		return
```
jgengo, line 17: you can reduce print + return by 

```python
sys.exit("message")
```

jgengo (approved):
Good job!

PR #165, opened 2020-04-06
my solution for chall04

saraarmiida:
I go through the grid starting from the bottom (n - y - 1) and if the character is '.' I check if in the position of (x, y + 1) there is a space and move it lower if so. By starting from the bottom i don't get the problem of sand stopping when theres more sand and an empty space after that sand.

### DIFF
--- a/chall04/spentti.py
+++ b/chall04/spentti.py
@@ -1,0 +1,21 @@
+#!/usr/bin/env python3
+
+print('input:')
+n = int(input())
+grid = []
+
+for row in range(n):
+	grid.append(list(input()))
+
+for y in range(n):
+	for x in range(n):
+		if grid[n - y - 1][x] == '.':
+			y2 = n - y - 1
+			while y2 + 1 < n and grid[y2 + 1][x] == ' ':
+				grid[y2][x] = ' '
+				grid[y2 + 1][x] = '.'
+				y2 += 1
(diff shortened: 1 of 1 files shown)


chall04/spentti.py
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env python3
+
+print('input:')
+n = int(input())
+grid = []
+
+for row in range(n):
+	grid.append(list(input()))
```
jgengo, line 8: can be reduced  the way:

```python
grid = [list(input()) for _ in range(n)]
```

Plus, you are not using "row" so you should move it as a blank identifier `_`.

chall04/spentti.py
```diff
@@ -0,0 +1,21 @@
...
+for y in range(n):
+	for x in range(n):
+		if grid[n - y - 1][x] == '.':
+			y2 = n - y - 1
+			while y2 + 1 < n and grid[y2 + 1][x] == ' ':
+				grid[y2][x] = ' '
+				grid[y2 + 1][x] = '.'
+				y2 += 1
+
+print('\noutput:')
```
jgengo, line 19: ? it's gonna be a problem for the diff output expected.

jgengo (approved):
Good job! It's super clean

PR #166, opened 2020-04-06
chall04 solution

shilena91:
- Create an 2D array from input (**arr**)
- Loop through the array,  if a character is `.` and the one below it is `space`,  move the character down
- After loop we have new array :)

### DIFF
--- a/chall04/hopham.py
+++ b/chall04/hopham.py
@@ -1,0 +1,19 @@
+#!/usr/bin/python3
+
+n = int(input())
+
+arr = []
+for j in range(n):
+    arr.append(list(input()))
+
+for s in range(n - 1):
+    for a, b in enumerate(arr[s]):
+        if b == '.':
+            if arr[s + 1][a] == ' ':
+                arr[s][a] = ' '
+                arr[s + 1][a] = '.'
+
+for string in arr:
+    for a in string:
+        print(a, end="")
+    print()


chall04/hopham.py
```diff
@@ -0,0 +1,19 @@
...
+    for a, b in enumerate(arr[s]):
+        if b == '.':
+            if arr[s + 1][a] == ' ':
+                arr[s][a] = ' '
+                arr[s + 1][a] = '.'
+
+for string in arr:
+    for a in string:
+        print(a, end="")
+    print()
```
d15ky, line 19: print(''.join(string))

chall04/hopham.py
```diff
@@ -0,0 +1,19 @@
...
+n = int(input())
+
+arr = []
+for j in range(n):
+    arr.append(list(input()))
+
+for s in range(n - 1):
+    for a, b in enumerate(arr[s]):
+        if b == '.':
+            if arr[s + 1][a] == ' ':
```
d15ky, line 12: you can combine those to:
`if b == '.' and arr[s + 1][a] == ' ':`
I feel like it won't damage readability but will make the code a little bit more pythonic

d15ky (approved):
A very clean and concise solution. It doesn't check for the wrong input tho :) Which is important. 
But overall I really liked the simplicity, good job here! 
I'd suggest reading about Pythonic way to write code, all the idioms and tricks you can use to make it even more simple and concise if you are interested (e.g. all the input reading could be done in arr = [list(input()) for _ in range(n)] )

PR #167, opened 2020-04-06
#04 - Built To Fall

tuommii:
Normalizing input to list of strings and then checking (starting from bottom) if swap is possible between " " and "."

### DIFF
--- a/chall04/mtuomine.py
+++ b/chall04/mtuomine.py
@@ -1,0 +1,57 @@
+#!/usr/bin/env python3
+# 3.6.9
+import sys
+
+def to_int(value):
+	try:
+		num = int(value)
+	except ValueError:
+		print("argument was wrong type")
+		sys.exit(1)
+	return num
+
+
+# Remove newlines and first row, append spaces to rows
+def normalize(grid, size):
+	# Feeling Go
+	for y, row in enumerate(grid):
+		row = row[:size]
+		grid[y] = row.replace("\n", "")
+		row_len = len(grid[y])
+		while row_len < size:
+			grid[y] += " "
+			row_len += 1
+	return grid[1:]
+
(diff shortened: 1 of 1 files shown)


chall04/mtuomine.py
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env python3
+# 3.6.9
+import sys
+
+def to_int(value):
+	try:
+		num = int(value)
+	except ValueError:
+		print("argument was wrong type")
+		sys.exit(1)
```
jgengo, line 10: you can use sys.exit argument to pass the print as parameter of sys.exit()

```python
sys.exit("argument was wrong type")
```

chall04/mtuomine.py
```diff
@@ -0,0 +1,57 @@
...
+				y -= 1
+		y = size
+		x -= 1
+	return grid[:size+1]
+
+
+def main():
+	grid = []
+	for line in sys.stdin:
+		grid.append(line)
```
jgengo, line 52: you can simplified that way:

```python
grid = sys.stdin.readlines()
```

chall04/mtuomine.py
```diff
@@ -0,0 +1,57 @@
...
+	size = to_int(grid[0])
+	print(*skyfall(normalize(grid, size), size - 1), sep='\n')
```
jgengo, line 54: wuw, that's nice!

chall04/mtuomine.py
```diff
@@ -0,0 +1,57 @@
...
+
+if __name__ == "__main__":
```
jgengo, line 56: good you protected it.

jgengo (approved):
Good job! That's clean and clear.

PR #168, opened 2020-04-06
Falling Sand mdesta chall04

mdesta:
1. Get the Input and Validate on both line size and if line contains only valid characters
2. Convert inputs into a simple string representation starting from bottom right and ending top-left . Example:
.#.
.##    ====> ..###..#.
#.. 
3. In the above representation, every nth element makes up a column from bottom-up
4. To Solve for a given Column, Start from bottom and recursively swap elements if current is empty and 
     next is sand

### DIFF
--- a/chall04/mdesta.py
+++ b/chall04/mdesta.py
@@ -1,0 +1,77 @@
+#!/usr/bin/env python3
+# Python3 Falling Sand Stone
+
+# To Reverse the string
+def revString(str):
+  return str[::-1]
+
+# PrintOut Error and Exit
+def printError():
+    print("Input Error")
+    exit()
+
+# Print final OutPut
+def printOutput():
+    i = 0
+    j = 0
+    while j < gridSize:
+        k = 0
+        while k < gridSize and i < len(finalString):
+            print(finalString[i], end ="")
+            i += 1
+            k += 1
+        print()
+        j += 1
+
+# Solves One Column Starting From given Column
+# Every nth (n = gridSize) element makes a column
+# Recurssively Swap elements if current is empty
+# and next is sand
+def solveColumn(rawString, kol):
+    i = kol
+    while i <= (len(rawString) - (2 * gridSize) + kol):    
+        if rawString[i] == ' ' and rawString[i + gridSize] == '.':
+            temp = rawString[i]
+            rawString[i] = rawString[i + gridSize]
+            rawString[i + gridSize] = temp
+            solveColumn(rawString, kol)
+        else:
+            i += gridSize
+    return rawString
+
+# Program Starts Here
+# Get grid size 
+gridSize = int(input())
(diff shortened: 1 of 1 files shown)


chall04/mdesta.py
```diff
@@ -0,0 +1,77 @@
...
+# Python3 Falling Sand Stone
+
+# To Reverse the string
+def revString(str):
+  return str[::-1]
+
+# PrintOut Error and Exit
+def printError():
+    print("Input Error")
+    exit()
```
jgengo, line 11: You can use `sys.exit("message")` instead of creating a function printError().

jgengo (approved):
Your solution is genuine and it seems good. However, I'm a bit confused by the code, it's a bit unclear.

For example while j < gridSize, could use range or enumerate python loop, etc.

Still, it seems to work fine.

mdesta:
> Your solution is genuine and it seems good. However, I'm a bit confused by the code, it's a bit unclear.
> 
> For example while j < gridSize, could use range or enumerate python loop, etc.
> 
> Still, it seems to work fine.

TL;DR
The idea is that a falling sand grid can be represented in 1D as a string.
Let's say Sand is A, space is B and Stone is C and consider a 3 by 3 grid as an example.
Starting from top-left and ending bottom-right, the below grid
AAA
BBB
CBC
Can be represented  by the string:
**AAABBBCBC**
For shortening the code, I'll work with the string reversed, and hence the string becomes:
**CBCBBBAAA**
We can see that for n=3 (grid size) every nth (3rd) element in the string makes a column.
We start solving from the last(right most) column and in the string, they are string[0], string[3] and string[6] from bottom up. The same for all other columns
Now we can recursively swap the characters in the string to solve for a given column.
1. Doing so for the last one column, the string representation will be (on the reverse order):
 AAA                                                                                                   AAB
 BBB     ==>     CBCBBBAAA     ==>     CBCABBBAA     ==>     BBA      
 CBC                                                                                                   CBC
2. Next for the middle column, the string representation will be:
 AAB                                                                                                   ABB
 BBA     ==>     CBCABBBAA     ==>     CACABBBBA     ==>     BBA      
 CBC                                                                                                   CAC
3. Finally for the first column, the string representation will be:
 ABB                                                                                                   BBB
 BBA     ==>     CACABBBBA     ==>     CACABABBB     ==>     ABA      
 CAC                                                                                                   CAC
4. Reversing the final string gives us **BBBABACAC** which is then easy to print out.

Therefore, what my program does is:
1.  Read the input into a string
2. Reverse the string so that it matches the above representation logic.
3. Change the string to a list for easy manipulation as strings are immutable in python.
4. Solve for every column using the above logic, i.e., while i < grideSize.
5. Once done, convert back the list to string and reverse it back for printing.
 .

PR #169, opened 2020-04-06
ptuukkan's solution

ptuukkan:
First validate the input. Then I chose to convert each "vertical line" to be a "horizontal line / string" in the program. That means that I can use split, join, and other functions to simulate the falling of sand. After I convert the lines back to original and print them.

### DIFF
--- a/chall04/ptuukkan.py
+++ b/chall04/ptuukkan.py
@@ -1,0 +1,54 @@
+#!/usr/bin/python3
+# Version 3.6.8
+
+import sys
+
+def validate_input(lines):
+	if len(lines) == 0:
+		return 0
+	dim = lines.pop(0)
+	if not dim.isnumeric():
+		return 0
+	dim = int(dim)
+	if len(lines) != dim:
+		return 0
+	for line in lines:
+		if len(line) != dim:
+			return 0
+		for c in line:
+			if not c in " .#":
+				return 0
+	return dim
+
+def convert_lines(lines, dim):
+	"Changes list lines from horizontal to vertical"
+	i = 0
+	converted_lines = []
+	while i < dim:
+		converted_lines.append("".join(line[i] for line in lines))
+		i += 1
+	return converted_lines
+
+def fall_sand(vlines):
+	"Simulates falling of sand for each line"
+	fallen_lines = []
+	for line in vlines:
+		i = 0
+		splitted = line.split('#')
+		for part in splitted:
+			splitted[i] = part.replace(" ", "").rjust(len(part), " ")
+			i += 1
+		fallen_lines.append("#".join(splitted))
+	return fallen_lines
+
+lines = []
+for line in sys.stdin.readlines():
+	lines.append(line.rstrip('\n'))
+dim = validate_input(lines)
+if dim == 0:
+	exit("Invalid input")
+vertical_lines = convert_lines(lines, dim)
+fallen_lines = fall_sand(vertical_lines)
+lines = convert_lines(fallen_lines, dim)
+for line in lines:
+	print(line)


chall04/ptuukkan.py
```diff
@@ -0,0 +1,54 @@
...
+	fallen_lines = []
+	for line in vlines:
+		i = 0
+		splitted = line.split('#')
+		for part in splitted:
+			splitted[i] = part.replace(" ", "").rjust(len(part), " ")
+			i += 1
+		fallen_lines.append("#".join(splitted))
+	return fallen_lines
+
```
d15ky, line 43: would be nice to protect this script from on import run (`if __name__ == '__main__':`)

chall04/ptuukkan.py
```diff
@@ -0,0 +1,54 @@
...
+lines = []
+for line in sys.stdin.readlines():
+	lines.append(line.rstrip('\n'))
```
d15ky, line 46: goood job! 
here is a shorter version to consider:
`lines = line.rstrip('\n') for line in sys.stdin.readlines()`

chall04/ptuukkan.py
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/python3
+# Version 3.6.8
+
+import sys
+
+def validate_input(lines):
+	if len(lines) == 0:
+		return 0
```
d15ky, line 8: would be nice to return, e.g. error message instead of 0 in case of an error

chall04/ptuukkan.py
```diff
@@ -0,0 +1,54 @@
...
+	dim = lines.pop(0)
```
d15ky, line 9: nice!

chall04/ptuukkan.py
```diff
@@ -0,0 +1,54 @@
...
+	if not dim.isnumeric():
+		return 0
+	dim = int(dim)
+	if len(lines) != dim:
+		return 0
+	for line in lines:
+		if len(line) != dim:
+			return 0
+		for c in line:
+			if not c in " .#":
```
d15ky, line 19: `if set(line) > set('#. '):`
set is very useful data type, I suggest learning it

chall04/ptuukkan.py
```diff
@@ -0,0 +1,54 @@
...
+				return 0
+	return dim
+
+def convert_lines(lines, dim):
+	"Changes list lines from horizontal to vertical"
```
d15ky, line 24: well done with function documenting!

chall04/ptuukkan.py
```diff
@@ -0,0 +1,54 @@
...
+	i = 0
+	converted_lines = []
+	while i < dim:
```
d15ky, line 27: `for i in range(dim):`

chall04/ptuukkan.py
```diff
@@ -0,0 +1,54 @@
...
+		converted_lines.append("".join(line[i] for line in lines))
```
d15ky, line 28: converted_lines = ["".join(line[i] for line in lines) for i in range(dim)]
the whole function in one readable line — that's the Python power ;)

chall04/ptuukkan.py
```diff
@@ -0,0 +1,54 @@
...
+		i += 1
+	return converted_lines
+
+def fall_sand(vlines):
+	"Simulates falling of sand for each line"
+	fallen_lines = []
+	for line in vlines:
+		i = 0
+		splitted = line.split('#')
+		for part in splitted:
```
d15ky, line 38: `for i, part in enumerate(splitted):`

chall04/ptuukkan.py
```diff
@@ -0,0 +1,54 @@
...
+			splitted[i] = part.replace(" ", "").rjust(len(part), " ")
```
d15ky, line 39: that's a nice Pythonic line

d15ky (approved):
Good job here! Clean function-based structure of main flow, some documentation on functions which is much appreciated, cool pythonic solutions. There is still a way to go in discovering Pythonic way to write Python code but I'm glad to see the signs that you are on the right way :) Well done.

PR #170, opened 2020-04-06
mmmm.. 

Rakutaa:
took first line of input for a prefix for grid. 
rest of the input made an array (grid).
checked error cases by len and regex
dropped the sand (was tricky since string seems to be immutable..)
printed.

### DIFF
--- a/chall04/vtran.py
+++ b/chall04/vtran.py
@@ -1,0 +1,37 @@
+#!/usr/bin/python3
+
+import re
+import sys
+
+siz = int(input())
+arr = []
+i = 0
+while i < siz:
+    arr.append(input())
+    i += 1
+i = siz - 1
+while i >= 0:
+    if (len(arr[i]) != siz) :
+        print("input not a grid")
+        sys.exit()
+    i -= 1
+listToStr = ''.join(map(str, arr))
+maximum = str(siz * siz)
+regex = "[. #]{" + maximum + "}"
+if not re.match(regex, listToStr) :
+    print("usage: [. #]*")
+    sys.exit()
+StrToList = list(listToStr)
+i = 0
+allcellsbutnotlastcolumns = siz * siz - siz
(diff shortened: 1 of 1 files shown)


chall04/vtran.py
```diff
@@ -0,0 +1,53 @@
...
+
+import re
+import sys
+
+def exitIfNotValidInput(siz, arr):
+    i = siz - 1
+    while i >= 0:
+        if (len(arr[i]) != siz) :
+            print("input not a grid")
+            sys.exit()
```
jgengo, line 12: You can pass the print inside the sys.exit()

```python
sys.exit("message")
```

chall04/vtran.py
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/python3
+#python 3.6.9
+
+import re
+import sys
+
+def exitIfNotValidInput(siz, arr):
+    i = siz - 1
+    while i >= 0:
```
jgengo, line 9: I would recommend using for in python instead of while.

chall04/vtran.py
```diff
@@ -0,0 +1,53 @@
...
+        print(arr[i])
+        i += 1
+
+def main ():
+    siz = int(input())
+    arr = []
+    i = 0
+    while i < siz:
+        arr.append(input())
+        i += 1
```
jgengo, line 49: a more pythonic way to write it would be:

```python
arr = [list(input()) for _ in range(n)]
```

chall04/vtran.py
```diff
@@ -0,0 +1,53 @@
...
+    exitIfNotValidInput(siz, arr)
+    done = droppingDots(arrToStr(arr), siz)
+    printter(done, siz)
```
jgengo, line 52: sys.exit(f"message {siz}")

jgengo (approved):
Good job!

PR #171, opened 2020-04-07
chall04

tangxing1220:
This challenge reminds me of fillit project. So there is no other options but two dimensional character array.  There are two main tasks.  to check/verify the input value. and there is 3 level of looping for "sand falling down". The important things are the criteria of looping. 

1, using python system function input() get input value from standard input(screen)
2, verifying the first input is one integer, give the warn info and exit if it is wrong.
3, verifying the length of input,  give the warn info and exit if it is wrong.
4, verifying the characters of input, give the warn info and exit if it is wrong.
5. there level of for looping:
   1st level loop: index is y.  It means the row. from the low to high
         2nd level loop: index is x. It means every character from the left to right
               3rd level loop: index is z. It means every row (y, from low to high)  "fall down" to ground. it 
                      it the last row.    
                compare the characters which are in the same column up and down.
6. print out the array.


chall04/xtang.py
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/python3
+
+import sys
+
+def main():
+    default = [" ", ".", "#"]
```
d15ky, line 6: "config" list — nice! easy to adjust if needed

chall04/xtang.py
```diff
@@ -0,0 +1,41 @@
...
+    world = []
+# get integer from standard input
+    try:
```
d15ky, line 9: Usage of try-except — great! good technique

chall04/xtang.py
```diff
@@ -0,0 +1,41 @@
...
+        size = int(input())
+    except:
```
d15ky, line 11: however, except clause is too broad. it's always better to specify particular error which you are expecting to handle (e.g., like ValueError)

chall04/xtang.py
```diff
@@ -0,0 +1,41 @@
...
+        print("Wrong number!")
+        sys.exit()
```
d15ky, line 13: sys.exit("Wrong number!")

chall04/xtang.py
```diff
@@ -0,0 +1,41 @@
...
+        print("Wrong size!")
+        sys.exit()
+# get input characters from standard input and verify the length and letters.    
+    for i in range(size):
+        world.append(input())
+        if len(world[i]) != size:
+            print("Wrong length!")
+            sys.exit()
+        for j in world[i]:
+            if not default.count(j):
```
d15ky, line 25: `if set(world[i]) > set(default):`
set is a very useful data type, suggest learning it

chall04/xtang.py
```diff
@@ -0,0 +1,41 @@
...
+                sys.exit()
+# compare two characters located at uppper postion and lower postion in the same column     
+    for y in reversed(range(size - 1)):
+        for x in range(size):
+            for z in range(size - 1 - y):
+                if world[y + z][x] == '.' and world[y + 1 + z][x] == ' ':
+                    world[y + z] = world[y + z][:x] + ' ' + world[y + z][x + 1:]
+                    world[y + 1 + z] = world[y + 1 + z][:x] + '.' + world[y + 1 + z][x + 1:]
+# print out the result
+    print("\n")
```
d15ky, line 36: hmm?

chall04/xtang.py
```diff
@@ -0,0 +1,41 @@
...
+    for k in world:
+        print(k)
+
+if __name__ == "__main__":
```
d15ky, line 40: well done with protection from on import run!

d15ky (approved):
Great job! It's a pleasure to read well-documented code. Solution is simple and nicely explained. 
I would suggest learning more about Pythonic way to write code (which means, among other things, less usage of nested loops and conditions). 
But anyway, that's a great improvement in Python shown here and I enjoyed reviewing your code.

PR #172, opened 2020-04-07
ran out of time...

MikeyLHolm:
I ran out of time and ran into problem of not being able to change strings as they aren't mutable. I guess changing them to ASCII or initially read them as something else than STR would be more optimal. 

It's lacking still character check and the actually fall of the sand isn't finished. But it was meant to be double loop replacing " " with "." and "." with " " on Y + 1.

### DIFF
--- a/chall04/mlindhol.py
+++ b/chall04/mlindhol.py
@@ -1,0 +1,34 @@
+#!/usr/bin/env python3
+# python 3.7.4 Windows 10
+
+import sys
+
+
+def input_error():
+    print("Bad input, n > 0 and only '.' ' ' and '#' allowed")
+    exit(-1)
+
+
+def main():
+    lines = []
+    lines = sys.stdin.read().splitlines()
+    n = int(lines[0])
+    if n < 1:
+        input_error()
+    lines.pop(0)
+
+    print("\n:::START POSITION:::\n")
+    print("N =", n)
+    for line in lines:
+        print(line)
+    print("--------------------")
+
+    for i in range(n):
+        for j in range(n):
+            if lines[i][j] is '.' and lines[i + 1][j] is ' ':
+                lines[i][j] = ' '
+                lines[i + 1][j] = '.'
+
+
+if __name__ == "__main__":
+    main()


mdesta:
I don't know if it is optimal, but I run into the same issue with strings being immutable, so my work around was to change it to a list and work on it, and finally join back list elements into a string.

chall04/mlindhol.py
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env python3
+# python 3.7.4 Windows 10
+
+import sys
+
+
+def input_error():
+    print("Bad input, n > 0 and only '.' ' ' and '#' allowed")
+    exit(-1)
```
jgengo, line 9: I would suggest avoiding creating a function for that and use sys.exit() with your message as argument of sys.exit(); 

chall04/mlindhol.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def main():
+    lines = []
+    lines = sys.stdin.read().splitlines()
```
jgengo, line 14:
```python
lines = sys.stdin.readlines()
```

chall04/mlindhol.py
```diff
@@ -0,0 +1,34 @@
...
+    n = int(lines[0])
+    if n < 1:
+        input_error()
+    lines.pop(0)
+
+    print("\n:::START POSITION:::\n")
+    print("N =", n)
```
jgengo, line 21: ?

chall04/mlindhol.py
```diff
@@ -0,0 +1,34 @@
...
+    print("--------------------")
+
+    for i in range(n):
+        for j in range(n):
+            if lines[i][j] is '.' and lines[i + 1][j] is ' ':
+                lines[i][j] = ' '
+                lines[i + 1][j] = '.'
+
+
+if __name__ == "__main__":
```
jgengo, line 33: good you protected it.

jgengo (commented):
Good job!

PR #173, opened 2020-04-09
add sadawi.rb ro chall05

salahadawi:
First, i check if a wikipedia article exists for the given argument. If not, a message is displayed and execution stops. Then, i read the article into memory. With regex, i find the first link, print the article name, and read that article next. Each article that has been visited is saved into an array to detect duplicates. This is then done in a loop until either the philosophy article is found, or a loop is detected.

### DIFF
--- a/chall05/sadawi.rb
+++ b/chall05/sadawi.rb
@@ -1,0 +1,33 @@
+#!/usr/bin/env ruby
+#ruby 2.5.1p57
+
+require 'open-uri'
+begin
+file = open('https://en.wikipedia.org/wiki/' + ARGV[0])
+rescue
+	puts "no wikipedia page for this word!"
+	exit
+end
+contents = file.read
+main = contents.match(/(?<=<p>)(?<!<\/p>)[\s\S]*?((?<=wiki\/)[a-zA-Z_()]*?(?="))[\s\S]*?<\/a>/)[1]
+visited = []
+count = 1
+puts "	Going to " + main + " (counter: " + count.to_s + ")"
+visited.push(main)
+
+while main != "Philosophy"
+	count += 1
+	file = open('https://en.wikipedia.org/wiki/' + main)
+	contents = file.read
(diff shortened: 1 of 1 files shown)


chall05/sadawi.rb
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env ruby
+#ruby 2.5.1p57
+
+require 'open-uri'
+begin
+file = open('https://en.wikipedia.org/wiki/' + ARGV[0])
```
jgengo, line 6: be careful of the indentation even if it;'s not important in Ruby, it's better to keep using it

chall05/sadawi.rb
```diff
@@ -0,0 +1,33 @@
...
+rescue
+	puts "no wikipedia page for this word!"
+	exit
+end
+contents = file.read
+main = contents.match(/(?<=<p>)(?<!<\/p>)[\s\S]*?((?<=wiki\/)[a-zA-Z_()]*?(?="))[\s\S]*?<\/a>/)[1]
+visited = []
+count = 1
+puts "	Going to " + main + " (counter: " + count.to_s + ")"
```
jgengo, line 15: you can use 

```ruby
puts "\tGoing to #{main} (counter: #{count})"
```

chall05/sadawi.rb
```diff
@@ -0,0 +1,33 @@
...
+visited.push(main)
```
jgengo, line 16: you can also write:
```ruby
visited << main
```

chall05/sadawi.rb
```diff
@@ -0,0 +1,33 @@
...
+
+while main != "Philosophy"
+	count += 1
+	file = open('https://en.wikipedia.org/wiki/' + main)
+	contents = file.read
+	main = contents.match(/(?<=<p>)(?<!<\/p>)[\s\S]*?((?<=wiki\/)[a-zA-Z_()]*?(?="))[\s\S]*?<\/a>/)[1]
+	if visited.include? main
```
jgengo, line 23: even if ruby allows to pass parameters that way I highly recommend keeping writing the other way.

jgengo (approved):
Good job!

PR #174, opened 2020-04-09
chall05 submit

Rouhija:
Two global variables, array of visited links and the initial argument which we can fall back to in case of errors.

- We recurse the function aristotle() until check_category initially stops the execution or we exit with CTRL+C (decided to catch that for fun)

- First we make an api call to wikipedia to check which categories the page belongs to. If we find philosophy in the json, we exit the program.

- If not, make a request to the link and add it to visited pages. Get the first link from the main text part with scan and ignore Help: etc links. Return the first correct link that has not been visited and call aristotle() again.

### DIFF
--- a/chall05/srouhe.rb
+++ b/chall05/srouhe.rb
@@ -1,0 +1,103 @@
+#!/usr/bin/ruby
+# ruby 2.5.1p57
+
+require 'json'
+require 'net/http'
+
+VISITED = []
+ARG = ARGV[0]
+
+
+def shut_down
+	puts "\nNo patience, huh?"
+	sleep 1
+end
+
+
+Signal.trap("INT") { 
+  shut_down 
+  exit
+}
+
+
+def get_first_link(res)
+	begin
+		i = res.body.index('<p>')
+		main = res.body.slice(i..-1)
+		links = main.scan(/(?<=<a href="\/wiki\/).[^"]*/)
+		links.each { |link|
+			if not link.include? ":"
+				if VISITED.include? link
+					puts "Skip %s, already seen this one" % link
+				else
+					return link
+				end
+			end
+		}
+	rescue
+		return ARG
+	end
+end
+  
+
+def make_request(page)
+	base = 'https://en.wikipedia.org/wiki/'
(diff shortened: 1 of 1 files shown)


Rouhija:
Noticed it's supposed to be the article philosophy instead of the page belonging to that category, so removed an unnecessary api call to check which categories a page belongs to.

chall05/srouhe.rb
```diff
@@ -0,0 +1,85 @@
...
+def shut_down
+	puts "\nNo patience, huh?"
+	sleep 1
+end
+
+
+Signal.trap("INT") { 
+  shut_down 
+  exit
+}
```
jgengo, line 20: why? if you would do something special like closing a database connection

chall05/srouhe.rb
```diff
@@ -0,0 +1,85 @@
...
+
+
+def get_first_link(res)
+	begin
+		i = res.body.index('<p>')
+		main = res.body.slice(i..-1)
+		links = main.scan(/(?<=<a href="\/wiki\/).[^"]*/)
+		links.each { |link|
+			if not link.include? ":"
+				if VISITED.include? link
```
jgengo, line 30: even if ruby allows it I highly recommend not passing parameters out of parenthesis. To keep it consistent with other languages you develop or will develop with.

chall05/srouhe.rb
```diff
@@ -0,0 +1,85 @@
...
+					puts "Skip %s, already seen this one" % link
```
jgengo, line 31: in Ruby you can write it that way

```ruby
puts "Skip #{link}, already seen this one"
```

chall05/srouhe.rb
```diff
@@ -0,0 +1,85 @@
...
+	else
+		puts "no wikipedia page for this word!: %s" % page
+		exit
+	end
+end
+
+
+def aristotle(arg, counter)
+	res = make_request(arg)
+	VISITED << arg
```
jgengo, line 59: good!

jgengo (approved):
Good job!

PR #175, opened 2020-04-09
Chall05 solution

tdawson-42:
To avoid using regex, I chose to use a HTML parser called Nokogiri to find hyperlinks in the source code. I'm limiting the parsing to only links that are within the "mw-content-text" div and within paragraph or list tags to only get links that are part of the main article body. I check for links within list tags so the parser doesn't get stuck on pages like [this](https://en.wikipedia.org/wiki/Corona) where there is no paragraphs.

When I have parsed all the links in an article, I iterate through them and take the first one that leads to another Wikipedia article and hasn't been seen yet (I add every visited page to a set). I split the URL path in the link to get the title of the Wikipedia article for printing it out, then recursively call the method with that page. When the page that will be opened next is "Philosophy" the program prints success message and exits.


chall05/tdawson.rb
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env ruby
+#ruby 2.5.1p57
+
+require 'open-uri'
+require 'nokogiri'
+require 'set'
+
+$WIKI_URL = String.new("https://en.wikipedia.org/wiki/")
+$seen_pages = Set.new
```
jgengo, line 9: Why using set ? I think a ruby array is enough, no?

chall05/tdawson.rb
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env ruby
+#ruby 2.5.1p57
+
+require 'open-uri'
+require 'nokogiri'
+require 'set'
```
jgengo, line 6: nice you used nokogiri!

chall05/tdawson.rb
```diff
@@ -0,0 +1,40 @@
...
+
+$WIKI_URL = String.new("https://en.wikipedia.org/wiki/")
+$seen_pages = Set.new
+
+def get_page_links(page)
+	file = open($WIKI_URL + page) rescue abort("no wikipedia page for this word!")
+	links = Nokogiri::HTML.parse(file)
+		.css('div#mw-content-text p a[href]', 'div#mw-content-text li a[href]')
+		.map { |elem| elem['href'] }
```
jgengo, line 15: really clean!

chall05/tdawson.rb
```diff
@@ -0,0 +1,40 @@
...
+def find_philosophy(page, counter = 1)
+	links = get_page_links(page)
+	next_page = next_wiki_page(links)
+	(puts "!!! Reach Philosophy !!!"; exit) if next_page.eql?("Philosophy")
+	$seen_pages << next_page;
+	puts "\tGoing to #{next_page} (counter: #{counter})"
+	find_philosophy(next_page, counter += 1)
+end
+
+if ARGV.empty? then start_page = "" else start_page = ARGV[0] end #start from main page if no arg given :D
```
jgengo, line 38: start_page = ARGV.empty? ? "" : ARGV[0] should works

jgengo (approved):
Good job!

chall05/tdawson.rb
tdawson-42: I used a set because I'm storing unique values (each seen page only needs to be included once) and from my understanding searching through a set is faster than through an array.

PR #176, opened 2020-04-09
chall05

osalmine:
First I check if the wikipedia page exists. After this I search the page starting from the div with the id "mw-content-text" because the page's main content starts there and so it ignores all links before that. I only search for links in <p> tags. It finds all links in those paragraphs and ignores the links with non-ascii values (sets nil in the place). After finding all of the links, skip the ones with nil and skip the ones that have already been visited. Set the url address for the next page to be the first new link found and print the message and check if the next page is Philosophy.

### DIFF
--- a/chall05/osalmine.rb
+++ b/chall05/osalmine.rb
@@ -1,0 +1,63 @@
+#!/usr/bin/env ruby
+
+# ruby version 2.6.3p62
+
+require 'nokogiri'
+require 'open-uri'
+require "net/http"
+
+str = ARGV[0]
+
+fail("Usage: enter one wikipedia page name") if ARGV.length() != 1
+
+url = URI.parse("https://en.wikipedia.org/wiki/#{str}")
+req = Net::HTTP.new(url.host, url.port)
+req.use_ssl = true
+res = req.request_head(url.path)
+
+if res.code.to_i == 404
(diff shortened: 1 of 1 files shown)


chall05/osalmine.rb
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env ruby
+
+# ruby version 2.6.3p62
+
+require 'nokogiri'
+require 'open-uri'
+require "net/http"
```
jgengo, line 7: yeah! really good you used nokogiri

chall05/osalmine.rb
```diff
@@ -0,0 +1,63 @@
...
+
+str = ARGV[0]
+
+fail("Usage: enter one wikipedia page name") if ARGV.length() != 1
```
jgengo, line 11: one line if is the most ruby way to write it! good job!

chall05/osalmine.rb
```diff
@@ -0,0 +1,63 @@
...
+
+url = URI.parse("https://en.wikipedia.org/wiki/#{str}")
+req = Net::HTTP.new(url.host, url.port)
+req.use_ssl = true
```
jgengo, line 15: good!

jgengo (approved):
Really clean code! 

Good jobb!

PR #177, opened 2020-04-09
chall05

vkuikka:
Checks if given wikipedia page exists and then opens second link from given page in a loop using regex until philosophy article is found.

### DIFF
--- a/chall05/vkuikka.rb
+++ b/chall05/vkuikka.rb
@@ -15,6 +15,7 @@ rescue
 end
 visited = []
 while (article != "Philosophy")
+	visited << article
 	file = open(url + article)
 	content = file.read
 	matches = content.scan(/(?<=(<p>)|(<h[0-9])).+?<a href=\"\/wiki\/([^\":]*)"/mu).drop(1)
@@ -26,7 +27,6 @@ while (article != "Philosophy")
 			puts "skip " + items[2] + " already visited"
 		end
 	end
-	visited << article
 	puts "	Going to " + article
 end
 puts "Found Philosophy!"
--- a/chall05/vkuikka.rb
+++ b/chall05/vkuikka.rb
@@ -1,0 +1,32 @@
+#!/usr/bin/ruby
+#ruby 2.5.5p157
+
+require 'open-uri'
+
+if (!article = ARGV[0])
+	return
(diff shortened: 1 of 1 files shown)


chall05/vkuikka.rb
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/ruby
+#ruby 2.5.5p157
+
+require 'open-uri'
+
+if (!article = ARGV[0])
+	return
+end
```
jgengo, line 8: always good to print a usage if there is a problem.

chall05/vkuikka.rb
```diff
@@ -0,0 +1,32 @@
...
+url = "https://en.wikipedia.org/wiki/"
+begin
+file = open(url + article)
```
jgengo, line 11: indentation issue here, even if it's not important in Ruby, always good to keep doing it

chall05/vkuikka.rb
```diff
@@ -0,0 +1,32 @@
...
+rescue
+	puts "No wikipedia page for this word"
+	return
+end
+visited = []
+while (article != "Philosophy")
+	visited << article
```
jgengo, line 18: nice!

chall05/vkuikka.rb
```diff
@@ -0,0 +1,32 @@
...
+	file = open(url + article)
+	content = file.read
+	matches = content.scan(/(?<=(<p>)|(<h[0-9])).+?<a href=\"\/wiki\/([^\":]*)"/mu).drop(1)
+	for items in matches
+		if !visited.include? items[2]
```
jgengo, line 23: even if Ruby allows it, I would recommend keeping writing () to pass parameters.

chall05/vkuikka.rb
```diff
@@ -0,0 +1,32 @@
...
+			article = items[2]
+			break
+		else
+			puts "skip " + items[2] + " already visited"
```
jgengo, line 27: in Ruby you can write
```ruby
puts "skip #{items[2]} already visited" 
```

jgengo (approved):
Good job.

jgengo:
Please, try to write more information into your PR description.

PR #178, opened 2020-04-09
rjaakonm chall05

rasmusjaa:
Use a loop with 2 functions (get source code and get the link) until Philosophy is found or article has no more new links.

Send program argument to a function that takes a string, concats it to Wikipedia URL, gets source code from URL and sends it to next function to find the first link.

Next function: get all paragraphs from source code with regex. Loop through each of them until first link is found. Links are found with regex by getting all links from one paragraph at once. If the first link is Philosophy we can finish, otherwise check that link is not already visited (if it is, the loop continues to next link/paragraph). If the link is not visited we go to the previous function again to get source code with the new link.


chall05/rjaakonm.rb
```diff
@@ -0,0 +1,54 @@
...
+require 'open-uri'
+
+$array = Array.new
+$counter = 0
+
+def getArticleSource(string)
+	url = 'https://en.wikipedia.org/wiki/'
+	source = open(url + string).read
+    rescue
+		puts 'no wikipedia page for this word!'
```
jgengo, line 13: there is a problem with your indent here

chall05/rjaakonm.rb
```diff
@@ -0,0 +1,54 @@
...
+	pattern = /<p>[\s\S]+?<\/p>/
+	paragraphs = source.scan(pattern)
+	if !paragraphs[0]
+		puts 'no wikipedia page for this word!'
+		exit
+	end
+
+    # Get wikipedia links from content and loop to find first link
+	pattern = /a href="\/wiki\/([\s\S]*?)["]/
+    paragraphs.each{ |paragraph|
```
jgengo, line 30: indent problem here again.

chall05/rjaakonm.rb
```diff
@@ -0,0 +1,54 @@
...
+		links = paragraph.scan(pattern)
+        links.each{ |link|
```
jgengo, line 32: if you are doing a multi-line each you should write it that way:

```ruby
links.each do |link|
  func1()
  func2()
end
```

chall05/rjaakonm.rb
```diff
@@ -0,0 +1,54 @@
...
+			if (link[0] == 'Philosophy')
+				puts '!!! Reach Philosophy !!!'
+				exit
+			elsif ($array[0] and $array.include? link[0])
```
jgengo, line 36: I suggest avoiding writing your params that way even If ruby allows it. It's better to keep consistent with other languages you develop or will develop with.

jgengo (approved):
Really clean code, good job!

I would suggest the Nokogiri library to parse HTML.

PR #179, opened 2020-04-09
Wikipage hopper

RandommUser:
Start by testing an argument is given. Define the base url. Initializing an array which will hold the page history. Make counters for steps taken and page number.

Main-loop: Check if on the right page. Check if visited this page already. Append to history if new. Call for the page reader

Page reader: Parse the page. Test if there is "this page does not exists" element in the page. Find the wanted link-elements using css-selectors and make them into a list.

Back in the main loop: Check if the page contains stuff we want. If no go back a step. If yes start the process again starting from the first link. Also regex match it is a "/wiki/"link as those are what we want to hop.


chall05/Gemfile_phakakos
```diff
@@ -0,0 +1,4 @@
+source "https://rubygems.org"
+
+gem 'nekogiri'
```
jgengo, line 3: 😂 nokogiri ?

chall05/phakakos.rb
```diff
@@ -0,0 +1,124 @@
+#!/usr/bin/env ruby
+# ruby version 2.6.6p146
+$VERBOSE = nil
+
+require 'HTTParty'
+require 'Nokogiri'
```
jgengo, line 6: good you used nokogiri!

chall05/phakakos.rb
```diff
@@ -0,0 +1,124 @@
...
+
+def base_url
+	"https://en.wikipedia.org"
+end
```
jgengo, line 10: better to create a const global variable then

chall05/phakakos.rb
```diff
@@ -0,0 +1,124 @@
...
+			if ($step > 0)
+				puts "Skip #{pname[-1]}, already seen this one"
+			end
+			$dup = 1
+			return()
+		else
+			if ($step > 0)
+				puts "\tGoing to #{pname[-1]} (counter: #{$step})"
+			end
+			$history.push(url)
```
jgengo, line 79: a more rubby way to write it is 

```ruby
$history << url
```

jgengo (approved):
Good job!

I like the fact you created class and use Nokogiri.

Be careful your gemfile isn't good :D

chall05/Gemfile_phakakos
RandommUser: Shieeet

PR #181, opened 2020-04-09
My implementations of Philosophy

DianaMukaliyeva:
1. First I checked if we have an argument.
2. I called a recursive function, that takes our given argument, array of already visited urls, counter and flag, that shows us whether we should skip the next link on this page.
3.  In this function, I checked whether wikipedia page exists, whether it is final_path, whether our link is part of Help_link (as I understand it, we should skip them), or whether we've already visited this page.
4. For finding exact link I have additional function in which I am using method "slice" a lot. I realize that I could do it better by using regex and maybe some other helpful methods.

### DIFF
--- a/chall05/dmukaliy.rb
+++ b/chall05/dmukaliy.rb
@@ -1,0 +1,49 @@
+#!/usr/bin/env ruby
+#ruby 2.5.1p57
+
+require 'open-uri'
+
+def go_to(url, visited, counter, seen)
+    start_url = "https://en.wikipedia.org/wiki/"
+    if url == "Philosophy"
+        puts "!!! Reach Philosophy !!!"
+        exit
+    end
+    begin
+        file = open(start_url.concat(url))
+    rescue
+        puts "no wikipedia page for this word!"
+        exit
+    end
+    content = file.read()
+    content = content.slice(content.index('<p>')..-1)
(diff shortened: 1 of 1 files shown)


chall05/dmukaliy.rb
```diff
@@ -0,0 +1,58 @@
...
+end
+
+def go_to(url, visited, counter, skip)
+    start_url = "https://en.wikipedia.org/wiki/"
+    begin
+        file = open(start_url.concat(url))
+    rescue
+        puts "no wikipedia page for this word!"
+        exit
+    end
```
jgengo, line 26: really good error handling.

chall05/dmukaliy.rb
```diff
@@ -0,0 +1,58 @@
...
+    if our_path.include? "Help:IPA/"
+        go_to(url, visited, counter, skip + 1)
+    end
+    if !visited.include? our_path
+        counter += 1
+        skip = 0
+        puts "\tGoing to " + our_path  + " (counter: " + counter.to_s + ")"
+        visited.push(our_path)
```
jgengo, line 40: a more ruby way to write it would be:

```ruby
visited << our_path
```

jgengo (approved):
Good job!

I would suggest trying Nokogiri lib to parse HTML it's really powerful.

PR #182, opened 2020-04-09
#05

tuommii:
### Logic
* I split HTML from first `<p>` tag.
* I'm interested only href-attributes that starts with `/wiki`
* The link is either new, seen, or the program is terminating

### DIFF
--- a/chall05/mtuomine.rb
+++ b/chall05/mtuomine.rb
@@ -1,0 +1,49 @@
+#!/usr/bin/env ruby
+
+#ruby 2.5.1p57
+
+require 'net/http'
+
+uri = URI("https://en.wikipedia.org/wiki/" + ARGV[0])
+counter = 0
+visited = []
+
+def traverse(visited, uri, counter)
+	html = Net::HTTP.get(uri)
+	if html.empty?
+		puts "no wikipedia page for this word!"
+		exit
+	end
+
+	content = html.split("<p>", 2)
+	linksArr = content[1].scan(/href\s*=\s*"([^"]*)"/)
+
+	# Bubblegum, couldn't combine with regex on tpo
+	res = []
+	linksArr.each { |link|
+		if link[0].match(/^\/wiki\//)
+			res.append(link[0])
+		end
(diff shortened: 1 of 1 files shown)


chall05/mtuomine.rb
```diff
@@ -0,0 +1,49 @@
...
+		puts "no wikipedia page for this word!"
+		exit
+	end
+
+	content = html.split("<p>", 2)
+	linksArr = content[1].scan(/href\s*=\s*"([^"]*)"/)
+
+	# Bubblegum, couldn't combine with regex on tpo
+	res = []
+	linksArr.each { |link|
```
jgengo, line 23: if you one-line a each in ruby you write it that way:

```ruby
linksArr.each { |link| puts link }
```

and if you want more lines:

```ruby
linksArr.each do |link|
  puts link
end
```

chall05/mtuomine.rb
```diff
@@ -0,0 +1,49 @@
...
+			res.append(link[0])
+		end
+	}
+
+	res.each { |link|
+		trimmed = link.split('/').last
+		if trimmed.eql? "Philosophy"
+			puts "!!! Reach Philosophy !!!"
+			exit
+		elsif visited.include? trimmed
```
jgengo, line 34: Ruby allows this kind of argument passed but I will always suggest keeping writing 
```ruby
visited.include?(trimmed)
```

To keep consistent with other languages you develop or will develop with.

chall05/mtuomine.rb
```diff
@@ -0,0 +1,49 @@
...
+			puts "Skip #{trimmed}, already seen this one"
+		else
+			puts "\tGoing to #{trimmed} (counter: #{counter})"
+			visited.append(trimmed)
```
jgengo, line 38: a more ruby way to write it:
```ruby
visited << trimmed 
```

jgengo (approved):
Good job!

I would suggest trying Nokogiri lib to parse HTML it's a powerful parser.

PR #183, opened 2020-04-09
Init road to philosophy

lapaset:
- Check that has the needed argument
- Create an array for visited sites
- While the page is not Philosophy loop:
   - net/http to get the contents of the page
   - If the response code is something else than 200 print the error message and exit
   - Starting from the first paragraph on the page loop until next unvisited site is found:
      - Find the next link
   - Update uri.path and add new page to visited

### DIFF
--- a/chall05/llahti.rb
+++ b/chall05/llahti.rb
@@ -1,0 +1,44 @@
+#!/usr/bin/ruby -w
+#ruby 2.3.7p456
+
+require 'net/http'
+
+if ARGV.length == 0
+    exit
+end
+
+uri = URI("https://en.wikipedia.org/wiki/#{ARGV[0]}")
+visited = [ARGV[0]]
+
+while uri.path != '/wiki/Philosophy'
+
+    res = Net::HTTP.get_response(uri)
+
+    if res.code != '200'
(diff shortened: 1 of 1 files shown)


chall05/llahti.rb
```diff
@@ -0,0 +1,44 @@
...
+        if visited.include?(nextPage)
+            puts "Skip #{nextPage}, already seen this one"
+        elsif !nextPage.start_with?("Help:IPA") && !nextPage.start_with?("File:")
+            break
+        end
+        startSearch = startSearch + startLink
+    end
+
+    uri.path = "/wiki/#{nextPage}"
+    visited << nextPage
```
jgengo, line 38: really ruby way to push into an array! nice

chall05/llahti.rb
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/ruby -w
+#ruby 2.3.7p456
+
+require 'net/http'
+
+if ARGV.length == 0
+    exit
```
jgengo, line 7: printing a usage would be nice

jgengo (approved):
Good job!

I would suggest using nokogiri lib to parse html

PR #184, opened 2020-04-09
ptuukkan's solution

ptuukkan:
The script's idea is to get the article's content and first match the area where to look for links to subsequent wiki articles. Then it gets an array of links using the scan method and moves to the first link. The script also stores visited pages in an array and selects another link if it has visited that one already.

### DIFF
--- a/chall05/ptuukkan.rb
+++ b/chall05/ptuukkan.rb
@@ -1,0 +1,46 @@
+#!/usr/bin/ruby
+#version: 2.5.1p57
+
+require 'net/http'
+
+def get_wiki(article)
+	net = Net::HTTP.new("en.wikipedia.org", 443)
+	net.use_ssl = true
+	response = net.get("/wiki/" + article)
+	if (response.code.match?(/^3\d\d$/) && response.header['location'])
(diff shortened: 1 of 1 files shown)


chall05/ptuukkan.rb
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/ruby
+#version: 2.5.1p57
+
+require 'net/http'
+
+def get_wiki(article)
+	net = Net::HTTP.new("en.wikipedia.org", 443)
+	net.use_ssl = true
```
jgengo, line 8: 👏good

chall05/ptuukkan.rb
```diff
@@ -0,0 +1,46 @@
...
+		puts "Skip " + URI.unescape(links[i][0]) + ", already seen this one"
+		i += 1
+	end
+	return links[i][0]
+end
+
+abort("Enter starting article as argument!") if ARGV.length != 1
+response = get_wiki(ARGV[0])
+visited = Array.new
+visited.push(ARGV[0])
```
jgengo, line 34: A more ruby way to write it 
```ruby
visited = []
visited << ARGV[0]
```
```

jgengo (approved):
Good job!

PR #185, opened 2020-04-10
Add krusthol's solution to Philosophy

krusthol:
Approach:
Start last minute before deadline, lookup Learn X in Y for Ruby. Lookup snippets and start testing, get results, be amazed. Skip learning the language, post something to challenge that often works in testing.

Code:
Object oriented mixed with C style whiles, taking advantage of Ruby's features such as array.last access and 'includes?'-feature. Assuming real content of wikipages start after first <p> tag, and assumes first <p>aragraph has enough links to proceed.

### DIFF
--- a/chall05/krusthol.rb
+++ b/chall05/krusthol.rb
@@ -4,7 +4,12 @@ require 'open-uri'
 
 counter = 0
 checked_pages = []
-starting_point = 'https://en.wikipedia.org/wiki/' + ARGV[0]
+if ARGV[0] != nil
+    starting_point = 'https://en.wikipedia.org/wiki/' + ARGV[0]
+else
+    puts "Road to Philosophy script usage: ./krusthol.rb [Wikipedia article]"
+    exit
+end
 begin
 contents = open(starting_point).read
 rescue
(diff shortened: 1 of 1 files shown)


chall05/krusthol.rb
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env ruby
+#ruby 2.5.1p57
+require 'open-uri'
+
+counter = 0
+checked_pages = []
+if ARGV[0] != nil
+    starting_point = 'https://en.wikipedia.org/wiki/' + ARGV[0]
+else
+    puts "Road to Philosophy script usage: ./krusthol.rb [Wikipedia article]"
```
jgengo, line 10: You should never hard-code your app name in the code. ( __FILE__ )

chall05/krusthol.rb
```diff
@@ -0,0 +1,47 @@
...
+    while road == nil do
+        road = split_p[i].scan(/(?<=<a href="\/wiki\/).[^"]*/)
+        i = i + 1
+    end
+    i = 0
+    while checked_pages.include?(road[i]) do
+        puts ("Skip " + road[i] + ", already seen this one")
+        i = i + 1    
+    end
+    checked_pages.push(road[i])
```
jgengo, line 43: I suggest a more ruby way to write it.

```ruby
checked_pages << road[i]
```

chall05/krusthol.rb
```diff
@@ -0,0 +1,47 @@
...
+counter = 0
+checked_pages = []
+if ARGV[0] != nil
+    starting_point = 'https://en.wikipedia.org/wiki/' + ARGV[0]
+else
+    puts "Road to Philosophy script usage: ./krusthol.rb [Wikipedia article]"
+    exit
+end
+begin
+contents = open(starting_point).read
```
jgengo, line 14: ident isn't important in Ruby, but always good to respect it.

jgengo (approved):
Good job!

I would suggest using Nokogiri lib ( https://nokogiri.org/ ) to parse the HTML.

PR #187, opened 2020-04-23
Add my solution sadawi.py

salahadawi:
A hacky solution for dealing with either standard input or files.
First, i check if any arguments have been given, and based on that, choose how to take input. I take the first line, split it and check if it only contains integers.
Then, i take each line from the file/input and add up the width of each book. By taking the largest shelf each time from the list of shelves, i will find the lowest amount of shelves needed to hold the total width of all books. 
All this is repeated again for each file given as an argument.

### DIFF
--- a/chall06/sadawi.py
+++ b/chall06/sadawi.py
@@ -14,6 +14,9 @@ def get_biggest(shelfs):
 if len(sys.argv) < 2:
 	lines = []
 	shelfs = input().split()
+	for n in shelfs:
+		if not n.isdigit():
+			sys.exit(sys.argv[0] + ": " + "stdin:" + " Can't read file")
 	while True:
 		try:
 			line = input()
@@ -40,6 +43,7 @@ else:
 	args = sys.argv[1:]
 	for arg in args:
 		lines = []
+		fail = 0
 		if len(args) > 1:
 			print(arg + ":")
 		try:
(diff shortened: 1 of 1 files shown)


chall06/sadawi.py
```diff
@@ -0,0 +1,84 @@
...
+def get_biggest(shelfs):
+		biggest = 0
+		for i in range(0, len(shelfs) - 1):
+			if int(shelfs[i]) > int(shelfs[biggest]):
+				biggest = i
+		size = shelfs[biggest]
+		del shelfs[biggest]
+		return int(size)
+
+if len(sys.argv) < 2:
```
jgengo, line 14: I would suggest having a `if __name__ == "__main__"

and add this protection into it.

https://stackoverflow.com/questions/419163/what-does-if-name-main-do

chall06/sadawi.py
```diff
@@ -0,0 +1,84 @@
...
+	lines = []
+	shelfs = input().split()
+	for n in shelfs:
+		if not n.isdigit():
+			sys.exit(sys.argv[0] + ": " + "stdin:" + " Can't read file")
```
jgengo, line 19: you should take a look how to format text + variable in python to avoid concatenation like that.

```python
sys.exit(f"{sys.argv[0]}: stdin: Can't read file")
```

chall06/sadawi.py
```diff
@@ -0,0 +1,84 @@
...
+		lines = []
+		fail = 0
+		if len(args) > 1:
+			print(arg + ":")
+		try:
+			file = open(arg, "r")
+		except:
+			print(sys.argv[0] + ": " + arg + ": Can't read file")
+			continue
+		shelfs = file.readline().split()
```
jgengo, line 54: not sure it has exactly the same behavior but you may take a look to the function readlines() 

https://www.tutorialspoint.com/python3/file_readlines.htm

jgengo (commented):
good job!

PR #188, opened 2020-04-23
Add solution

hiiri:
Check command line argument count to determine if reading is done from files or from standard input.
If files, for each file given - try to read it and print an error if there is no read permission or the file doesn't exist. Standard input has the same approach from here on. Read first line of file to get the shelves and their widths, then read the rest of the lines and get the width from each book. Sort the widths of the shelves to compare them to the total size of the books combined. If current shelf size isn't enough, add the next biggest shelf size to the total shelf size. If all shelves are still too small in total, print error for not enough space.

### DIFF
--- a/chall06/alcohen.py
+++ b/chall06/alcohen.py
@@ -45,7 +45,7 @@ def shelves(arg):
         print(sys.argv[0] + ": " + arg + ": Can't read file")
         return
     
-    shelves.sort(reverse=True)
+    shelves.sort(key=int, reverse=True)
     shelves_total = 0
     for i in range(len(shelves) - 1):
         if (shelves[i].isdigit()):
--- a/chall06/alcohen.py
+++ b/chall06/alcohen.py
@@ -1,0 +1,69 @@
+#!/usr/bin/env python3
+
+import sys
+from os import path
+import errno
+
+def parse_content(content):
+    total = 0
+    for i in content:
+        if (i):
+            width = i.split()[0]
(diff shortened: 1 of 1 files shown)


chall06/alcohen.py
```diff
@@ -0,0 +1,69 @@
...
+    if (args > 0): #read from file
+        for arg in sys.argv[1:]:
+            if (args > 1):
+                print(arg + ":")
+            shelves(arg)
+    else: #read from command line
+        shelves("stdin")
+
+if __name__=="__main__":
+    main()
```
jgengo, line 69: that's clean.

You could have directly written your main inside this one while it's only called here.

chall06/alcohen.py
```diff
@@ -0,0 +1,69 @@
...
+    
+    shelves.sort(key=int, reverse=True)
+    shelves_total = 0
+    for i in range(len(shelves) - 1):
+        if (shelves[i].isdigit()):
+            shelves_total += int(shelves[i])
+        if (total <= shelves_total):
+            print(i + 1)
+            return
+    print(sys.argv[0] + ": " + arg + ": Not enough space in the given shelves")
```
jgengo, line 56: You should take a look at printing format to avoid concatenation

```python
print(f"{sys.argv[0]}: {arg}: Not enough space in the given shelves")
```

jgengo (commented):
Really good and clean code.

PR #190, opened 2020-04-23
rjaakonm chall06

rasmusjaa:
In main check if the program was given parameters. If not, read lines from std in, otherwise loop all args and read files line by line. Give error message if a file can't be read or is not valid.

Send all lines to a function that first checks the first line. It must be only integers and at least one integer. I sort it into descending order to get the biggest shelves first. 

Other lines are split by spaces into array per line. The array must have integer first and at least something else as the name of the book (can't be just width). Book widths are put into an int array and that array is also ordered in descending order. Shelves are filled like this:
1. Put the biggest book to the biggest shelf. Delete that book (it's width) from book array and remove the same amount of width from the biggest shelf width. Start again from book index 0.
2. Test if now the biggest book fits into the first shelf. If not, test the next book until not even the smallest fits.
3. Go to the next shelf and do the same, fill it with the biggest book left first and go to the next book that fits.
4. Do this until all books are in shelves or no shelves left.
4. If all books are deleted from book array we are done and we return how many shelves we used (index of shelf for loop + 1 at that moment). If we come out of shelf loop before book array is empty we return error.

This is quite effective way to fill the shelves, but not completely perfect. It can handle this case perfectly:
20 10 10
15 book a
10 book b
5 book c

But because it fits the biggest book first into the biggest shelf it thinks that this case needs 3 shelves instead of putting books b and c to shelf with width 30:
30 20 20
20 book a
15 book b
15 book c


chall06/rjaakonm.py
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/python3
+
+import sys
+import fileinput
+
+def solve(lines, name, files):
+	if len(lines) < 1: 
+		print(sys.argv[0], ": ", name, ": Can't read file", sep = "")
```
jgengo, line 8: you may take a look to format string with variable.

```python
print(f"{sys.argv[0]}: {name}: Can't read file")
```

chall06/rjaakonm.py
```diff
@@ -0,0 +1,74 @@
...
+			lines.append(line)
+		solve(lines, 'stdin', 1)
+	else:
+		iterargs = iter(sys.argv)
+		next(iterargs)
+		for arg in iterargs:
+			lines = []
+			try:
+				for line in fileinput.input(arg):
+					lines.append(line)
```
jgengo, line 66: I think this could be one-lined

chall06/rjaakonm.py
```diff
@@ -0,0 +1,74 @@
...
+	print(sys.argv[0], ': ', name, ': Not enough space in the given shelves', sep = '')
+
+def main():	
+	lines = []
+	if len(sys.argv) == 1:
+		for line in fileinput.input():
+			lines.append(line)
+		solve(lines, 'stdin', 1)
+	else:
+		iterargs = iter(sys.argv)
```
jgengo, line 60: most of your variables and functions have the snake_case type, I would suggest using it for all of them so `iterargs` -> `iter_args`

jgengo (commented):
Good job!

PR #191, opened 2020-04-23
Bookself calculator

RandommUser:
Pretty basic but flexible reader.

First it checks if the program got any arguments. If yes then for-loop thru them all, reading the lines.If no then start reading stdin, which is saved into an array. Splice the first line, "atoi" them and error if not only numbers. Then read the following lines. The width is first in each line so we splice them again and take the first argument,  try to turn it into an int and add it to the total required width.

Once completed we start to figure out how many shelfs we need. Find the widest shelf, add it to the total and remove it from the array. Continue this loop till we have space to fit the books or when we run out of shelfs. Then print the amount or error given the result.

Program prints out the file name in errors or when dealing with multiple files, as required, and adds a linebreak between multiple file result with very simple check.

### DIFF
--- a/chall06/phakakos.py
+++ b/chall06/phakakos.py
@@ -6,11 +6,10 @@ import fileinput
 def errnoHandle(code, source):
         #replace with (sys.argv[0] + " ") for flexible use. My enviroment does not match the tests, which gives a different out put, so I'm just gonna hardcode it :^)
         beginning = "./phakakos.py "
-        if code == 1:
+        if code == -1:
                 print(beginning + source + ": Can't read file")
-        elif code == 2:
+        elif code == -2:
                 print(beginning + source + ": Not enough space in the given shelve")
-        return (code)
 
 def readLoop(lines, name):
         totalW = 0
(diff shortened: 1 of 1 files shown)


RandommUser:
So apparently the bookshelves have sidewalls. So, I made a loop which places the books into the shelves, starting from the largest available and using already used shelves when fitting. It is not the most thorough, but should work in most cases. Added tests for to fit the largest books in the collection as well as total space available for purchase.

chall06/phakakos.py
```diff
@@ -0,0 +1,114 @@
...
+import sys
+import fileinput
+
+def errnoHandle(code, source):
+        #replace with (sys.argv[0] + " ") for flexible use. My enviroment does not match the tests, which gives a different out put, so I'm just gonna hardcode it :^)
+        beginning = "./phakakos.py "
+        if code == -1:
+                print(beginning + source + ": Can't read file")
+        elif code == -2:
+                print(beginning + source + ": Not enough space in the given shelves")
```
jgengo, line 12: nice!

how comes your environment does not match the tests? What was the output of your environment? and what is your environment?

jgengo (commented):
Your readLoop is quite tough to read. I think you could have split a little bit more some of your actions in your method by clearer name.

Also, I highly suggest reading the PEP8: 
https://www.python.org/dev/peps/pep-0008/#function-and-variable-names

your functions and variables names are not following the PEP8.

chall06/phakakos.py
RandommUser: I used Windows and run it thru the Python Shell so it gives me the full path to the file rather than the "./name.py". It should give the correct output in Unix, but left it hardcoded for now.

PR #192, opened 2020-04-23
ptuukkan's solution

ptuukkan:
First validate the input and then calculates the minimum amount of shelves needed to store all the books by calculating total number of books and then subtracting shelf capacity starting from the largest shelf.

### DIFF
--- a/chall06/ptuukkan.py
+++ b/chall06/ptuukkan.py
@@ -1,0 +1,69 @@
+#!/usr/bin/env python3
+
+import sys
+
+def validate(lines):
+	shelves = lines[0].split(' ')
+	if any([not shelf.isnumeric() for shelf in shelves]):
+		return "", "", False
+	collections = [line.split(' ')[0] for line in lines[1:]]
+	if not collections or any([not collection.isnumeric() for collection in collections]):
+		return "", "", False
+	shelves = [int(i) for i in shelves]
+	collections = [int(i) for i in collections]
+	return shelves, collections, True
+
+def read_contents(source, stdin):
+	if (stdin):
+		try:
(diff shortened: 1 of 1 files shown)


chall06/ptuukkan.py
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/env python3
+
+import sys
+
+def validate(lines):
+	shelves = lines[0].split(' ')
+	if any([not shelf.isnumeric() for shelf in shelves]):
```
jgengo, line 7: nice!

chall06/ptuukkan.py
```diff
@@ -0,0 +1,69 @@
...
+		return "", "", False
+	collections = [line.split(' ')[0] for line in lines[1:]]
```
jgengo, line 9: nice!

chall06/ptuukkan.py
```diff
@@ -0,0 +1,69 @@
...
+	if not collections or any([not collection.isnumeric() for collection in collections]):
```
jgengo, line 10: nice!

chall06/ptuukkan.py
```diff
@@ -0,0 +1,69 @@
...
+			print(f"{file}:")
+			fit_books(sys.argv[0], file, False)
+			if args > 2:
+				print("")
+			args -= 1
+	else:
+		fit_books(sys.argv[0], "stdin", True)
+
+if __name__ == "__main__":
+	main()
```
jgengo, line 69: as main is only used here, you could have directly add it here.

jgengo:
Good job!

PR #194, opened 2020-04-23
solution

mashalink:
I check if any arguments have been given, and based on that, choose how to take input.
I try to read it and print an error if there is no read permission or the file doesn't exist. 
Read first line of file to get the shelfs and their length, then read the rest of the lines and get the length from each book. If length of biggest shelf is smaller than the length of the biggest book - print error.   
I start to fill the biggest shelf with the books, one by one, starting with the biggest book. If next book does not fit into the shelf it goes the next shelf. Each next book I try to put into the biggest shelf, if it is not possible, I go down to the smaller shelfs.

### DIFF
--- a/chall06/mlink.py
+++ b/chall06/mlink.py
@@ -1,0 +1,79 @@
+#!/usr/bin/env python3
+import fileinput
+import sys
+
+def error(error_message, filename):
+    print(f'{sys.argv[0]}: {filename}: {error_message}')
+
+def solution(lines, filename):
+
+    book = []
+    bookshelfs = [int(i) for i in lines[0].split()]
+
+    if len(bookshelfs) == 0:
+        error("Can\'t read file", filename)
+        return
+    for i in range(1, len(lines)):
+            lenght = int(lines[i].split()[0])
+            book.append(lenght)
+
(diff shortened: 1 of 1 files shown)


chall06/mlink.py
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env python3
+import fileinput
+import sys
+
+def error(error_message, filename):
+    print(f'{sys.argv[0]}: {filename}: {error_message}')
```
jgengo, line 6: I'm not sure creating a function error for that is really useful.

would have been better to use a built-in function like for example sys.stderr.write()

chall06/mlink.py
```diff
@@ -0,0 +1,79 @@
...
+                j += 1
+                flag = True
+                break
+            else:
+                i += 1
+        if i > min_bookshelfs:
+            min_bookshelfs = i
+        if flag == False:
+            error('Not enough space in the given shelves', filename)
+            exit()
```
jgengo, line 47: exit can take a print argument. so you would have save a line here

jgengo (commented):
clean. good job!

PR #195, opened 2020-04-24
Add krusthol's solution to Bookshelf Problem

krusthol:
Approach:
Coding Python again, so frequent googling and Learn X in Y cheat sheet and other materials guiding the process of course. The resulting code sprawled off to be messy, I coded this one with a sub-optimal overall structure... But it seems to work, and the standard input seems to be working snappily as well. I made it usable with the "<< EOF" method of the examples, with manual standard inputting from the terminal by running just the pure executable, and also piping with | operator seemed to work. Most time spent went to error handling and the various modes for input to be processed and toying around the environment, testing things a bit more thoroughly than usual, as I flunked on the last challenge due to an inputting issue I had not even tested.

Code:
I came up with the puzzle logic of sorting the shelves so that you always want to take the biggest shelves into use over the smaller ones. In my solution you want to start filling it with the biggest books first, thus also sorting the books is done before starting. Then we loop through the possibilities of filling the shelves with the books, while detecting always, if no book could anymore fit into any free shelf so we can do aborting with FIT_ERR then. I did not give the puzzle aspect that deep thought, it solves the example cases in similar minimum amount of shelf units as the examples show, with this fairly simple double iteration logic, so I assume it works to an extent.

### DIFF
--- a/chall06/krusthol.py
+++ b/chall06/krusthol.py
@@ -4,20 +4,20 @@ import sys
 READ_ERR = "Can't read file"
 FIT_ERR = "Not enough space in the given shelves"
 
-def proceed_with_error(many_args, stdin_mode, index, message):
+def print_proceed(many_args, std_input, index, message):
 	if many_args:
 		print(f"{sys.argv[index]}:")
-	if stdin_mode:
+	if std_input:
 		print(f"{sys.argv[0]}: stdin: {message}")
 	else:
 		print(f"{sys.argv[0]}: {sys.argv[index]}: {message}")
 	return (index + 1)
 
 def book_fitter(index):
-	many_args = True if len(sys.argv) > 2 else False
-	stdin_mode = True if len(sys.argv) == 1 else False
+	many_args = len(sys.argv) > 2
+	std_input = len(sys.argv) == 1
 	books = []
-	if stdin_mode:
+	if std_input:
 		lines = []
 		while True:
 			try:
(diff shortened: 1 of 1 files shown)


chall06/krusthol.py
```diff
@@ -0,0 +1,86 @@
...
+		while True:
+			try:
+				std_line = input()
+				if std_line == '':
+					break
+				lines.append(std_line)
+			except EOFError:
+				break
+		if not lines:
+			return (print_proceed(many_args, std_input, index, READ_ERR))
```
jgengo, line 31: quite special... :D

chall06/krusthol.py
```diff
@@ -0,0 +1,86 @@
...
+			while j < len(filled_shelves):
+				if int(books[i]) <= filled_shelves[j]:
+					filled_shelves[j] -= int(books[i])
+					books[i] = 'X'
+					break
+				j += 1
+			i += 1
+		if 'X' not in books:
+			return (print_proceed(many_args, std_input, index, FIT_ERR))
+		books[:] = (value for value in books if value != 'X')
```
jgengo, line 73: I guess you are trying to add into a python list online value in books that are not X?

books[:] = (...) 
can probably written: 
books = [(...)]

chall06/krusthol.py
```diff
@@ -0,0 +1,86 @@
...
+		shelves = lines[0].split()
+		lines.pop(0)
+		for line in lines:
+			new_book = line.split()[0]
+			books.append(new_book)
+	elif not std_input:
+		try:
+			with open(sys.argv[index]) as open_file:
+				shelves = open_file.readline().split()		
+				line = open_file.readline()
```
jgengo, line 41: not sure it has exactly the same behaviors

but I think readline().split() may be written readlines()

jgengo (commented):
Good job!

PR #196, opened 2020-04-24
chall06 solution

shilena91:
- If file cannot read, print error.
- I created 2 int arrays, 1 array is for bookshelves and another is for books
- If any shelf or book is not number, print error
- Iterate 2 arrays in function `get_shelf`, it returns number of shelves if books can fit them, otherwise return 0.
- Print output based on the return from `get_shelf`

### DIFF
--- a/chall06/hopham.py
+++ b/chall06/hopham.py
@@ -1,0 +1,69 @@
+#!/usr/bin/python3
+
+import sys
+import string
+
+def get_shelf(shelves, books):
+    i = 0
+    while i < len(shelves) :
+        if len(books) == 0:
+            return (i)
+        for b in books:
+            if b <= shelves[i]:
+                shelves[i] -= b
+                books.remove(b)
+                if len(books) == 0:
+                    break
+                if shelves[i] != 0:
+                    i = -1
+                break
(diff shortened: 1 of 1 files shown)


chall06/hopham.py
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/python3
+
+import sys
+import string
+
+def get_shelf(shelves, books):
+    i = 0
+    while i < len(shelves) :
```
jgengo, line 8: this looks really similar to the Hive C code.

In python you maybe use `for value in shelves:`, it's more pythonic way to write your code, and avoid you having a i, etc.

chall06/hopham.py
```diff
@@ -0,0 +1,69 @@
...
+        try:
+            file = open(i)
+
+            lines = file.readlines()
+            line1 = lines.pop(0)
+            shelves = line1.split()
+            for s in shelves:
+                if s.isdigit() is False:
+                    print("%s: %s: Can't read file"% (sys.argv[0], i))
+                    return
```
jgengo, line 35: you can use : exit() with an argument will print and exit.

PR #197, opened 2020-04-24
Pulls

DianaMukaliyeva:
Approach used: sorted shelves and books by width from maximum to minimum. Then looped through books and in every iteration looped through shelves. subtracted  current book width from the current shelf and continue looping through books and looking if they are fit in shelf.

### DIFF
--- a/chall06/dmukaliy.py
+++ b/chall06/dmukaliy.py
@@ -5,31 +5,46 @@ import sys
 def error(error_message, filename):
     print(f'{sys.argv[0]}: {filename}: {error_message}')
 
-def get_collections_size(data):
-    collections = []
+def get_sorted_books_width(data):
+    books = []
     for i in range(1, len(data)):
-        collection = data[i]
-        collections.append(int(collection.split()[0]))
-    return collections
+        book = data[i]
+        books.append(int(book.split()[0]))
+    books.sort(reverse=True)
+    return books
+
+def get_amount_shelves(shelves, books):
+    shelves_amount = -1
+    for i in range(0, len(books)):
+        found = False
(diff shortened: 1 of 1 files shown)


chall06/dmukaliy.py
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/python3.7
+
+import sys
+
+def error(error_message, filename):
+    print(f'{sys.argv[0]}: {filename}: {error_message}')
```
jgengo, line 6: you could have use exit() with argument, but not sure you really need to create your own function for handling error while it only print a line.

chall06/dmukaliy.py
```diff
@@ -0,0 +1,72 @@
...
+    if len(shelves) == 0:
+        exit()
+    books = get_sorted_books_width(data)
+    if sum(books) == 0:
+        print('0')
+        return
+    shelves.sort(reverse=True)
+    if shelves[0] < books[0]:
+        error('Not enough space in the given shelves', filename)
+        return
```
jgengo, line 42: yeah so if you would have use exit() you would have save a line here, and two for the function.

chall06/dmukaliy.py
```diff
@@ -0,0 +1,72 @@
...
+                found = True
+                break
+        if j > shelves_amount:
+            shelves_amount = j
+        if not found:
+            return -1
+    return shelves_amount + 1
+
+def parse_and_solve(data, filename):
+    shelves = [int(shelf) for shelf in data[0].split()]
```
jgengo, line 32: nice!

jgengo (commented):
good job!

PR #198, opened 2020-04-24
llahti fit the books

lapaset:
1. start by reading the input
 -> decided to have separate functions for reading file and stdin
2. parse first input line to shelves variable that is an ascending list of shelves
3. parse rest of the file to collection variable that is an ascending list of books
4. solve:
  - get the minimum amount of shelves needed for the total width of collection
  - use recursion to try and fit the collection to the amount of shelves: because we might want to buy books in the future too we always take the widest shelves
  - if a book fits, but next ones don't take the book off the shelve and try it to the next as long as possible
  - if the collection is empty so all the books have fit return the result, if not add a shelve as long as theres some left
5. print the result
6. if there is more than 1 file, start from beginning with the next one

### DIFF
--- a/chall06/llahti.py
+++ b/chall06/llahti.py
@@ -1,0 +1,120 @@
+#!/usr/bin/env python3
+#Python 3.8.2
+import sys
+import os
+
+def errorString(path, error):
+    format = "{}: {}: {}"
+    return format.format(sys.argv[0], path, error)
+
+def startCount(shelves, collection):
+    if shelves[0] < collection[0] or sum(collection) > sum(shelves):
+        return -1
+    count = 1
+    shelveSum = shelves[0]
+    while sum(collection) > shelveSum:
+        shelveSum += shelves[count]
+        count += 1
+    return count
+
+def fill(shelves, collection):
+    if not collection:
+        return True
+    i = 0
+    while i < len(shelves):
+        if collection[0] < shelves[i]:
+            shelves[i] -= collection[0]
+            return fill(shelves, collection[1:])
+        i += 1
+    return False
+
+def solve(path, shelves, collection):
+    collection.sort(reverse=True)
+    count = startCount(shelves, collection)
+    if count == -1:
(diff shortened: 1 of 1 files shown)


chall06/llahti.py
```diff
@@ -0,0 +1,123 @@
+#!/usr/bin/env python3
+#Python 3.8.2
+import sys
+import os
+
+def errorString(path, error):
+    format = "{}: {}: {}"
+    return format.format(sys.argv[0], path, error)
```
jgengo, line 8: I think it was not useful to create your own errorString function.

I also suggest to read the PEP8 python: https://www.python.org/dev/peps/pep-0008/#function-and-variable-names

It recommend snake_case function and variable names

chall06/llahti.py
```diff
@@ -0,0 +1,123 @@
...
+        line = sys.stdin.readline()
+
+    return solve(path, shelves, collection)
+
+def main():
+    args = len(sys.argv)
+
+    if args < 2:
+        if sys.stdin.isatty():
+            print('Usage:\t./llahti.py <collectionfile.txt>\n   or:\t./llahti.py <collection as stdin>')
```
jgengo, line 99: sys.argv[0] 

don't hardcore your app name into your usage.

chall06/llahti.py
```diff
@@ -0,0 +1,123 @@
...
+def readStdin():
+    path = 'stdin'
+    shelves = getShelves(sys.stdin.readline(), True)
+    if not shelves:
+        return errorString(path, "Can't read file")
+    collection = []
+    line = sys.stdin.readline()
+    while line:
+        try:
+            collection.append(int(line.split().pop(1)))
```
jgengo, line 87: I think all of this could have been one-lined with readlines() instead of readline() but you would probably not been able to use try/except, and would have to parse your lines after.

jgengo (commented):
Good job!

PR #199, opened 2020-04-25
Add ohakola's solution

hakolao:
Sorry for late solution. Didn't realise there had come out a new challenge :(

After parsing file or stdin & validating input:

Algorithm:
1. Sort shelves from largest to smallest (the larger a shelve is, the less we need space)
2. For each book index, and for each shelf index:
    - `shelves_needed`: Run a while loop consuming books. If shelf count goes over available shelves, abort while loop. Iteration starting from each book index and shelf index gets tested.
    - On success (each book was placed), update result if it's smaller than existing result.

Result is then returned & printed.

This solution will get slow on large inputs, so I don't think it's very efficient. But seems to work here.

### DIFF
--- a/chall06/ohakola.py
+++ b/chall06/ohakola.py
@@ -1,33 +1,34 @@
 #!/usr/bin/python3
 import sys
 
-# Loops through one round of books and tries to fit them to shelves
+# Loops through books and tries to fit them to shelves
 # If count goes too high, return -1
 def shelves_needed(books, shelves, shelf_index, book_index):
 	count = 1
 	curr_shelf = shelves[shelf_index]
 	books_len = len(books)
 	shelves_len = len(shelves)
-	i = book_index
-	j = shelf_index
-	for k in range(books_len):
-		if curr_shelf >= books[i]:
-			curr_shelf = curr_shelf - books[i]
-			i += 1
-			if i >= books_len:
-				i = 0
+	books_placed = 0
+	while books_placed < books_len:
+		if curr_shelf >= books[book_index]:
+			curr_shelf = curr_shelf - books[book_index]
+			book_index += 1
+			if book_index >= books_len:
+				book_index = 0
+			books_placed += 1
 		else:
 			count += 1
 			if count > shelves_len:
 				return -1
-			j += 1
-			if j >= shelves_len:
-				j = 0
-			curr_shelf = shelves[j]
+			shelf_index += 1
+			if shelf_index >= shelves_len:
+				shelf_index = 0
+			curr_shelf = shelves[shelf_index]
 	return count
 
 # Sorts shelves from largest to smalles after which attempt shelves_needed
-# function. For each book index and for each shelf index try to find solution
+# function for each book index and for each shelf index
+# returns best solution
 def variable_shelf_pack(books, shelves, program, source):
 	MAX_INT = 999999999
 	result = MAX_INT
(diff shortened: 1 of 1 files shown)


chall06/ohakola.py
```diff
@@ -0,0 +1,83 @@
...
+		lines = sys.stdin.readlines()
+	else:
+		print(f"{program}: Too many args, only one file accepted")
+		exit()
+	try:
+		shelves = lines.pop(0).split(" ")
+		for i, s in enumerate(shelves):
+  			shelves[i] = int(s.rstrip())
+	except:
+		print(f"{program}: {source}: Invalid shelf input given")
```
plantran, line 73: Nice!

chall06/ohakola.py
```diff
@@ -0,0 +1,83 @@
...
+			print(f"{program}: {source}: No such file")
+			exit()
+	elif arg_len == 1:
+		lines = sys.stdin.readlines()
+	else:
+		print(f"{program}: Too many args, only one file accepted")
+		exit()
+	try:
+		shelves = lines.pop(0).split(" ")
+		for i, s in enumerate(shelves):
```
plantran, line 70: Another way to approach this in a more "pythonic" way would be:
```
shelves = [ int(s.rstrip()) for s in shelves ]
```
It's called [list comprehension](https://docs.python.org/3/tutorial/datastructures.html#list-comprehensions).

chall06/ohakola.py
```diff
@@ -0,0 +1,83 @@
...
+			count += 1
+			if count > shelves_len:
+				return -1
+			shelf_index += 1
+			if shelf_index >= shelves_len:
+				shelf_index = 0
+			curr_shelf = shelves[shelf_index]
+	return count
+
+# Sorts shelves from largest to smalles after which attempt shelves_needed
```
plantran, line 29: As your comment describes the purpose of your function overall, you should prefer documenting it with a docstring. On the other hand, you can add comments to more specific parts of the function that could not be obvious. For example, you could briefly comment on your usage of the `MAX_INT`.

chall06/ohakola.py
```diff
@@ -0,0 +1,83 @@
...
+	while books_placed < books_len:
+		if curr_shelf >= books[book_index]:
+			curr_shelf = curr_shelf - books[book_index]
+			book_index += 1
+			if book_index >= books_len:
+				book_index = 0
+			books_placed += 1
+		else:
+			count += 1
+			if count > shelves_len:
```
plantran, line 21: One other pythonic way to approach this would be to create your custom error (e.g. `ShelvesCountError` - I'm terrible at naming), raise it and catch it in the `variable_shelf_pack()` function.

chall06/ohakola.py
```diff
@@ -0,0 +1,83 @@
...
+	source = "stdin"
+	arg_len = len(sys.argv)
+	if arg_len == 2:
+		source = sys.argv[1]
+		try:
+			with open(sys.argv[1]) as f:
+				lines = f.readlines()
+		except:
+			print(f"{program}: {source}: No such file")
+			exit()
```
plantran, line 62: `exit()` can take a string to print directly on STDERR. In that case, you would only have to write:
```
exit(f"{program}: {source}: No such file")
```

plantran:
Hi!

Unfortunately, you submitted your file after the last deadline so I will close this pull request. Nonetheless, this is a good piece of code, good job! 👍

hakolao:
Thanks for the comments. All great!

I’ll try not to miss the deadlines on the next ones :)

PR #200, opened 2020-04-27
chall07/vkuikka.py

vkuikka:
Keeps the distance to the edge of the array saved and increments it to avoid printing already printed characters. Four loops iterate array right, down, left and up each time increasing the distance to the edges. If array length is not even, the middle character is printed separately.

### DIFF
--- a/chall07/vkuikka.py
+++ b/chall07/vkuikka.py
@@ -9,6 +9,14 @@ def main():
 	y = 0
 
 	del(nums[0])
+	for arg in nums:
+		for char in arg:
+			if (char < '0' or char > '9'):
+				print("usage: " + sys.argv[0] + " <1-9 squared_rows...>")
+				return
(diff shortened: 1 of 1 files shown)


chall07/vkuikka.py
```diff
@@ -0,0 +1,46 @@
...
+	if length == 0:
+		print("usage: " + name + " <1-9 squared_rows...>")
+		return
+	for arg in nums:
+		for char in arg:
+			if (char < '0' or char > '9'):
+				print("usage: " + name + " <1-9 squared_rows...>")
+				return
+		if (len(arg) != length):
+			print("usage: " + name + " <1-9 squared_rows...>")
```
jgengo, line 22: instead of print + return you can use exit("message")

jgengo (commented):
good job

PR #201, opened 2020-04-27
Add solution to chall07

hiiri:
Checks that input is square first, then puts the input into a list of lists. Traverse the "square" list of lists by popping the first list (top), then the last element of each list (right), then the last list (bottom), finally the first element of each list (left). Repeat until nothing is left to be iterated upon.

### DIFF
--- a/chall07/alcohen.py
+++ b/chall07/alcohen.py
@@ -1,0 +1,46 @@
+#!/usr/bin/env python3
+
+import sys
+import itertools
+
+if __name__ == "__main__":
+    array = []
+    args = len(sys.argv) - 1
+    if (args == 0):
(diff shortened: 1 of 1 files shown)


chall07/alcohen.py
```diff
@@ -0,0 +1,46 @@
...
+
+import sys
+import itertools
+
+if __name__ == "__main__":
+    array = []
+    args = len(sys.argv) - 1
+    if (args == 0):
+        print(f"usage: {sys.argv[0]} <1-9 squared_rows...>")
+        exit()
```
jgengo, line 11: instead of print + exit, you can pass your exit message as an argument of exit.

PR #202, opened 2020-04-27
add solution sadawi.py

salahadawi:
A weird way to solve this problem, i start by surrounding the input with 'a' characters. I then start moving through the input clockwise, saving the numbers to an array and converting saved numbers from the input into 'a' characters, so i know which numbers have already been visited. Each time a wall is encountered (character that is not a number), the direction of movement rotates clockwise. Once we are completely surrounded by walls and cannot move, the sort is done and results are printed.

### DIFF
--- a/chall07/sadawi.py
+++ b/chall07/sadawi.py
@@ -1,0 +1,72 @@
+#!/usr/bin/env python3
+
+import sys
+
+if __name__=="__main__":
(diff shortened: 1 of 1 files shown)


jgengo (commented):
nothing to mention. Good job!

Quite unique way to solve it. I'm curious to see the trace.

PR #203, opened 2020-04-27
chall07

Rouhija:
1. Validate rows and input, error catching in main()
2. Format input to 2d array
3. while loop until all slots are visited
    - 4 loops (for directions to traverse) after each loop, increment or decrement start and end for x and y so we wont loop same x, y coord twice


jgengo (commented):
Really clean code.

Your own error handling, really well done.

PR #204, opened 2020-04-27
submit chall07

tdawson-42:
We start by checking for invalid input:
* number of arguments (rows) is not within the specified range (1-20)
* the input contains non-digit characters
* the length of each given row doesn't match the total number of rows (since we need _n_ rows of _n_ length for a square)

 If input is valid we then create a matrix from the input. In a loop, we always append the top row and then the right-most column (representing the snail moving to the right and then down) to the path-variable. We then slice the matrix removing the top row and right-most column and rotate the matrix 180 degrees so that the bottom row becomes the top row and start over. At the end we print the path with String,join.

### DIFF
--- a/chall07/tdawson.py
+++ b/chall07/tdawson.py
@@ -5,7 +5,7 @@ import numpy as np
 
 rows = len(sys.argv) - 1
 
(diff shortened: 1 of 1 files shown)


jgengo (commented):
the shortest code so far. 

nice usage of numpy.

PR #205, opened 2020-04-27
ptuukkan's solution

ptuukkan:
First validates the input, then iterates on the array, printing top line, right vertical line, bottom line reversed, left vertical line reversed and deleting printed elements. Stops when whole array is processed.

### DIFF
--- a/chall07/ptuukkan.py
+++ b/chall07/ptuukkan.py
@@ -1,0 +1,51 @@
+#!/usr/bin/env python3
+
+import sys
+
+def print_usage(progname):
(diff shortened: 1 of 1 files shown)


chall07/ptuukkan.py
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env python3
+
+import sys
+
+def print_usage(progname):
+	print(f"usage: {progname} <1-9 squared_rows...>")
+	exit()
+
```
jgengo, line 8: instead of print + exit, you can use only exit("message")

jgengo (commented):
clean code. Good job!

PR #206, opened 2020-04-27
chall07

rasmusjaa:
NEW SOLUTION:
My first solution was too heavy and I had a good idea of what I wanted to do instead so I optimized it. Now it's fast and simple:
1. Check that input is valid (has 1-20 arguments, argument amount is same as all argument lengths and all arguments are digits)
2. First variable is square size (length/width), second is 0. Loop through arguments with 4 for loops using those 2 variables and append all characters to list, at the end of each loop minus 1 and plus 1 from variables. Stop when second variable is same or bigger then first.
3. Print string.

OLD SOLUTION (only in first commits):
Loop all arguments to list array. If there are no arguments or length of an argument is different than the number of arguments or char in an argument is not 0-9, give usage.

While list array has elements, loop them in a function that appends first row elements into another array, appends the last character of each middle row, appends last row elements in reverse and appends the first character of each middle row in reverse. Append middle row only if rows is more than 3 and first characters if there is more than 1 column. Last row is appended if rows is more than 2. Middle row characters are removed after appending, last and first row in the end of append loop to keep append loop simple.

Row and column size calculations and column checks could be removed since input validation makes sure that they are always the same, but this method could now be used for other than square shapes too as long as rows are equal length.

When all rows all deleted from original list array we print new array with separator " ,".


jgengo (commented):
really clean and understandable solution. 

Good job

PR #207, opened 2020-04-27
My solution

HandsomeTom:
I have 3 functions, 2 dealing with error checking and exiting. The third one, make_final_print, is the actual solver which will be called as many times as necessary. It adds to the final string the "outline" of the remaining box (startpos argument). If the box has an odd number of elements the edge case is checked before continuing (are we in the final elements in the middle, if yes, just return that)

### DIFF
--- a/chall07/tmaarela.py
+++ b/chall07/tmaarela.py
@@ -1,0 +1,39 @@
+import sys
+
+def error_msg():
+	print("usage: ./tmaarela.py <1-9 squared_rows...>")
+	quit()
+
+def error_check(args):
+	if (len(args) <= 0):
+		error_msg()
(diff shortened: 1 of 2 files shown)


chall07/tmaarela.py
```diff
@@ -0,0 +1,39 @@
+import sys
+
+def error_msg():
+	print("usage: ./tmaarela.py <1-9 squared_rows...>")
+	quit()
```
jgengo, line 5: you can reduce print + quit with exit("message")

jgengo (commented):
Sadly, I've to refuse this PR you've submitted an extra file.

https://github.com/hivehelsinki/remote-challs/pull/207/files

PR #208, opened 2020-04-27
A lot of list comprehension

lkallio:
We just append everything according to the rules. I use list comprehension and [:::] (whatever it's called) to make life easier. Probably does more operations than is necessary, but it's neat enough for me.


jgengo (commented):
clean sys.argv[1:] etc is called "slicing" :)

Good job!

PR #209, opened 2020-04-27
Solution #07

tuommii:
Cutting edges until nothing left


jgengo (commented):
nothing to mention. It's clear, I can see you are using the Slice Slicing :) I can see Golang has changed your way to code.

PR #210, opened 2020-04-27
snail-challenge by sluhtala for python3

sluhtala:
Hello!
It is maybe a bit messy but makes sense to me. First reading the arguments and some error handling. I store every number in one-dimensional nums-array. new_nums is where I store the snail-array. I created a function that appends only the outer edge of the square using edge variable. And I loop that until there are no edge left. (x is the position in nums-array) Finally I print all numbers from new_nums with comma.


chall07/sluhtala.py
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/python3
+import sys
+
+def put_usage():
+    usage = "usage: ./sluhtala.py <1-9 squared_rows...>"
+    print(usage)
+    exit()
```
jgengo, line 7: don't hardcode your login use sys.argv[0] instead.

Plus you can use exit argument to avoid using print.

chall07/sluhtala.py
```diff
@@ -0,0 +1,52 @@
...
+        x -= size
+        new_nums.append(nums[x])
+    edge -= 1
+    return int(edge)
+
+if __name__ == "__main__":
+    args = sys.argv
+    if len(args) < 2:
+        put_usage
+    args.pop(0)
```
jgengo, line 34: You can also use `del(sys.argv[0])`, but .pop(0) is good!

chall07/sluhtala.py
```diff
@@ -0,0 +1,52 @@
...
+         if args[i][j].isnumeric() == 0:
+              put_usage()
+         nums.append(args[i][j])
+    edge = size
+    new_nums = []
+    x = 0
+    while edge > 0:
+        edge = print_sqr(nums, x, edge, new_nums)
+        x += size
+        x += 1
```
jgengo, line 51: x += (size + 1)

jgengo (commented):
Good job

PR #211, opened 2020-04-27
Hardcode hero

RandommUser:
Sooo yea...

Basically the solution logic is you reading "square" in the following order: top, right, bottom, left. So I just made a while loop to read thru it all, with a specific exit for even-numbered squares. Cheeky instructions show the even number being different but they didn't use any in the output examples, almost missed that.

Nothing else is really special, print has extra tag to prevent new line. The usage uses arg[0] for the name.

Thought about doing it all in the argv strings and counters, but that seemed a more ugly  solution so this was the better alternative

### DIFF
--- a/chall07/phakakos.py
+++ b/chall07/phakakos.py
@@ -8,6 +8,8 @@ def usage():
 
 def input_test():
         height = len(sys.argv) - 1
+        if height > 20:
+                usage()
         for arg in sys.argv[1:]:
                 # is good length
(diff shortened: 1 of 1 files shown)


jgengo (commented):
nice solution.

I don't have much to say it's quite clear. I would only suggest avoiding comment your code when it's not necessary for example:

#is only numeric
if arg.isnumeric() == False:

PR #212, opened 2020-04-27
My solution for chall07

DianaMukaliyeva:
Approach used: arguments were converted to a 2-dimensional array.
I am going through array in the loop and  do next:
    1. Save the first row to the result array and remove this row from the original array. 
    2. Append the last column to the result array and remove it from the original array.
    3.Append the last row in reverse order to the result array and remove this row from the original array.
    4. Append the first column in reverse order to the result array and remove this column from the original array.
I repeat the loop until there is no element left in the array.

### DIFF
--- a/chall07/dmukaliy.py
+++ b/chall07/dmukaliy.py
@@ -1,0 +1,35 @@
+#!/usr/bin/python3.7
+
+import sys
+
+def check_validity(data):
+    if not data or len(data) > 20:
+        sys.exit(f'usage: {sys.argv[0]} <1-9 squared_rows...>')
+    for row in data:
+        if len(row) != len(data) or not row.isnumeric():
+            sys.exit(f'usage: {sys.argv[0]} <1-9 squared_rows...>')
+
+def solve(data):
+    arr = [list(row) for row in data]
+    result = []
+
+    while arr:
+        result.extend(arr.pop(0))
+        for row in arr:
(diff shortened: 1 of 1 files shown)


chall07/dmukaliy.py
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/python3.7
+
+import sys
+
+def check_validity(data):
```
jgengo, line 5: you may have notice some functions/methods are called
isalpha, isnumeric, etc.

So better to keep the same logic with your method name: 

You could have name is is_valid(data):

but good still :)

chall07/dmukaliy.py
```diff
@@ -0,0 +1,35 @@
...
+    if not data or len(data) > 20:
+        sys.exit(f'usage: {sys.argv[0]} <1-9 squared_rows...>')
```
jgengo, line 7: Good!

chall07/dmukaliy.py
```diff
@@ -0,0 +1,35 @@
...
+    for row in data:
+        if len(row) != len(data) or not row.isnumeric():
+            sys.exit(f'usage: {sys.argv[0]} <1-9 squared_rows...>')
+
+def solve(data):
+    arr = [list(row) for row in data]
```
jgengo, line 13: nice!

jgengo (commented):
Good job!

PR #213, opened 2020-04-27
initial commit challenge 07

mdesta:
The basic logic is:
1.  Print first raw of the Matrix
2. Remove first raw
3. If there is more than one row
4. -> Rotate Counter ClockWise by 90 Degree
5. Set the number of rows using the new matrix
6. Repeat until there is no more row to print

### DIFF
--- a/chall07/mdesta.py
+++ b/chall07/mdesta.py
@@ -1,0 +1,52 @@
+#!/usr/bin/env python3
+# Python3 snake
+
+import sys
+
+# Rotate Matrix by 90 Degrees Counter-ClockWise
+def rotate_matrix( matrix ):
+    return [[matrix[j][i] for j in range(len(matrix))] for i in range(len(matrix[0])-1,-1,-1)]
+
+# Print Only first raw of Matrix
+# Adds a ", " if there is more row to print 
+def printTop(arr, rows):
+    i = 0
+    for j in range(len(arr[i])):
+        print(arr[i][j], end='')
+        if rows > 1:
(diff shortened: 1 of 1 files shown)


chall07/mdesta.py
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env python3
+# Python3 snake
+
+import sys
+
+# Rotate Matrix by 90 Degrees Counter-ClockWise
+def rotate_matrix( matrix ):
+    return [[matrix[j][i] for j in range(len(matrix))] for i in range(len(matrix[0])-1,-1,-1)]
```
jgengo, line 8: damn! I luv it!

chall07/mdesta.py
```diff
@@ -0,0 +1,52 @@
...
+# Set the number of rows using the new matrix
+# Repeat
+def main():
+    name = str(sys.argv[0])
+    matrix = sys.argv
+    matrix.pop(0)
+    rows = len(matrix)
+    if validateInput() or len(sys.argv) < 1:
+        print("usage: " + name + " <1-9 squared_rows...>")
+    else:
```
jgengo, line 43: you can skip indent with a return or exit
```python
if validate(...):
  exit("...")

while rows > 0:
```

It's not like a big deal, but it makes code more readable when you have more lines than that :)

jgengo (commented):
Nice comments on top of your methods. You can also use multiline comments.

Good job!

PR #215, opened 2020-04-27
Chall07 / Snail Sort solution

mashalink:
I check if a square is valid - should be range from a size of 1*1 to a size of 20*20, number of chars in each row should be the same as number of rows in total. And I also check that there are only chars from 1 to 9 in use. If something not valid I return error and exit the programme. 

As my script must print all of the elements in each row of the array arranged from outermost elements to the middle element, traveling clockwise, I start while loop until rows are over. Before starting each of the next 3 following steps I check if there are still rows.

1. I save each first row in a "newarray" without changes and remove in "array" and reduce the number of rows by one. 
2. I save all the chars from the end of each row to a "newarray" and remove them in "array"; Then I save a last row in a reverse order and reduce a number of rows by one. 
3. I save all the first chars from each row going from the last to the first row.

If I there are still left rows I start from the step 1.
While printing the array I add comma and space after each char.

### DIFF
--- a/chall07/mlink.py
+++ b/chall07/mlink.py
@@ -1,0 +1,46 @@
+#!/usr/bin/env python3
+import sys
+
+rows = len(sys.argv) - 1
+array = []
+newarray = []
+
+if 1 <= rows <= 20:
+
+    for i in range(1, rows + 1):
+        if len(sys.argv[i]) == rows:
+            s = sys.argv[i]
+            for j in range(rows):
+                if s[j] > '9' or s[j] < '0':
+                    print(f"usage: {sys.argv[0]} <1-9 squared_rows...>")
+                    exit() 
+            array.append(list(sys.argv[i]))
+        else:
+            print(f"usage: {sys.argv[0]} <1-9 squared_rows...>")
+            exit()
+
+    while rows > 0:
+
+        if rows > 0:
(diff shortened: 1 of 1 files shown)


chall07/mlink.py
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env python3
+import sys
+
+rows = len(sys.argv) - 1
+array = []
+newarray = []
+
+if 1 <= rows <= 20:
```
jgengo, line 8: I did not even know that was possible in python o_O

chall07/mlink.py
```diff
@@ -0,0 +1,46 @@
...
+
+    for i in range(1, rows + 1):
+        if len(sys.argv[i]) == rows:
+            s = sys.argv[i]
+            for j in range(rows):
+                if s[j] > '9' or s[j] < '0':
+                    print(f"usage: {sys.argv[0]} <1-9 squared_rows...>")
+                    exit()
```
jgengo, line 16: instead of print + exit, you can directly give the message into the exit argument. 

exit("message")

jgengo (commented):
Good job, I would recommend reading : https://stackoverflow.com/questions/419163/what-does-if-name-main-do

PR #216, opened 2020-04-28
rpehkone

vossr:
Check input for errors according to rules, then print spiral as a square that gets smaller.


chall07/rpehkone.py
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/python3
+
+import sys
+
+def usage():
+	print("usage: " + sys.argv[0] + " <1-9 squared_rows...>")
+	exit()
```
jgengo, line 7: instead of print + exit, you can directly give exit an argument message to print. 

exit("your message")

jgengo (commented):
Good job.

I think some of your while-loop could have been split into few functions with clearer method name to understand what's happing.

PR #217, opened 2020-04-28
Init snail

lapaset:
1. check that there are arguments and that they are formatted correctly
  - if not print usage and return
2. print the strings from outside in using while-loop, size of the square and index i
  - if the size is uneven, check if at the end after the top row
  - else printing ends after bottom row of the i == size / 2 - 1 round

### DIFF
--- a/chall07/llahti.py
+++ b/chall07/llahti.py
@@ -1,0 +1,53 @@
+#!/usr/bin/env python3
+#Python 3.8.2
+import sys
+import math
+
+def snail(lines):
+    size = len(lines[0])
+    i = 0
+    while (i < size / 2):
+        if size % 2 == 1 and i == math.floor(size / 2):
+            print(lines[i][i:size-i])
+            return
+        
+        line = list(lines[i][i:size-i])
+        for c in line:
+            print(c, end=', ')
+        j = i + 1
+        while (j < size - i - 1):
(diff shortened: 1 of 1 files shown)


chall07/llahti.py
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env python3
+#Python 3.8.2
+import sys
+import math
+
+def snail(lines):
+    size = len(lines[0])
+    i = 0
+    while (i < size / 2):
+        if size % 2 == 1 and i == math.floor(size / 2):
```
jgengo, line 10: why do you need math.floor() here? does int(size/2) wouldn't be enough?

I'm not sure

chall07/llahti.py
```diff
@@ -0,0 +1,53 @@
...
+            print(lines[size-1-i][k] + ',', end=' ')
+            k -= 1
+        if (i + 1 < size / 2):
+            print(lines[size-1-i][k] + ',', end=' ')
+        else:
+            print(lines[size-1-i][k])
+            return
+        while (j > i + 1):
+            j -= 1
+            print(lines[j][i] + ',', end=' ')
```
jgengo, line 32: This is a bit messy, and could probably split into few functions with clearer method name.

jgengo (commented):
The code is clear, but I think the logic part could be improved to be more readable.

Good job!

PR #218, opened 2020-04-28
chall07 solution

shilena91:
I just iterate all elements of 2D array in clockwise.
The function do that take parameters which are number of element in array, length of each element, and the array itself.

### DIFF
--- a/chall07/hopham.py
+++ b/chall07/hopham.py
@@ -49,4 +49,3 @@ for i in arr:
 		exit(f"usage: {sys.argv[0]} <1-9 squared_rows...>")
 
 spiralPrint(arr_len, arr_len, arr)
-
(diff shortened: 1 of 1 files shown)


chall07/hopham.py
```diff
@@ -0,0 +1,51 @@
...
+			l += 1
+
+	for i in arr[:-1]:
+		print(i, end= ", ")
+	print(arr[-1])
+
+
+if len(sys.argv) < 2:
+	exit(f"usage: {sys.argv[0]} <1-9 squared_rows...>")
+arr = sys.argv[1:]
```
jgengo, line 45: try to use meaningful variable names.

here for example: args = sys.argv[1:] would have been better.

chall07/hopham.py
```diff
@@ -0,0 +1,51 @@
...
+arr_len = len(arr)
```
jgengo, line 46: here argc :)

jgengo (commented):
Good job!

The spiral is really clear and your comments help a lots.

PR #219, opened 2020-04-28
tkuumola solution chall07

T7Q:
1. Input validation: 
- number of arguments > 1 and < 21
- input consists of digits only
- length of each argument to make sure we receive the square

2. Solution: 
- in the while loop restricted by the total number of numbers, append the numbers to result string with for loops following the snail pattern (right, down, left, up)
- print result

### DIFF
--- a/chall07/tkuumola.py
+++ b/chall07/tkuumola.py
@@ -1,0 +1,54 @@
+#!/usr/bin/env python3
+
+import sys
+
+def main():
+    length = len(sys.argv)
+    if length < 2 or length > 21:
+        sys.exit("usage: " + sys.argv[0] + " <1-9 squared_rows...>")
+
+    for i in range (1, length):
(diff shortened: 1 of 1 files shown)


chall07/tkuumola.py
```diff
@@ -0,0 +1,54 @@
...
+                break
+            result.append(input_array[last_row][i])
+            k+=1
+        last_row -= 1
+        for i in range (last_row, first_row - 1, -1):
+            if (k >= size):
+                break
+            result.append(input_array[i][first_col])
+            k+=1
+        first_col += 1
```
jgengo, line 49: This whole part could have been broken into small functions with meaningful method name. to make it a little bit clearer, or some comments.

jgengo (commented):
really good. I can see a lot of improvements in your way to code python since the challenge has started.

PR #220, opened 2020-04-28
final v

MikeyLHolm:
First validating that 0 < n < 21 & input is a square with digits only. Then i loop thru the list by doing all the 4 sides each loop. Appending the right numbers to a new list and then removing them from the original, thus making the rectangle smaller.

Is a messy and lengthy one and it probably could be done without max/min index with just slicing or something. Somehow I didn't get all the things on one-liners that I wanted ( pops and all the other list.append+manipulation ) as they were providing me errors.

Windows gives me bit different cat -e on Git Bash, hope all is well on your system:
```
$ ./mlindhol.py "123" "456" "789" | cat -e
1, 2, 3, 6, 9, 8, 7, 4, 5^M$
```

### DIFF
--- a/chall07/mlindhol.py
+++ b/chall07/mlindhol.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 # python 3.8.2 Windows 10
 
 import sys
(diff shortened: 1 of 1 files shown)


jgengo (commented):
I have not so much to say, your code is really good, break into small functions. 

Good job mlindhol !

PR #221, opened 2020-05-04
add solution sadawi.c

salahadawi:
I use an index variable to go through the string one character at a time, checking manually if I've reached a combination of either "ei" or "ie".
Then, i check to see if i can safely access the character right before this combination, by checking that we are not at index 0. If not at index 0, i check the previous character before the combination, and if that character breaks the spelling rule, i return 0.
Otherwise, once the whole string has been checked,  i return 1.

### DIFF
--- a/chall08/sadawi.c
+++ b/chall08/sadawi.c
@@ -1,0 +1,24 @@
+
+#include <string.h>
+
+int	ft_ie_except_after_c(char* str)
+{
+	int i;
+
+	i = 0;
+	while (str[i])
+	{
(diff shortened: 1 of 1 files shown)


tuommii (approved):
Nice, very clear code! I like the fact that you are checking `ie` or `ei` first and only then `"c"`.

chall08/sadawi.c
```diff
@@ -0,0 +1,24 @@
+
+#include <string.h>
```
tuommii, line 2: Ups =)

PR #222, opened 2020-05-04
chall08

rasmusjaa:
For loop through the string. The string is ok by default (return 1) except in 2 cases (return 0):
1. We find 'c' followed by 'i' followed by 'e'.
2. We find 'e' followed by 'i'. If 'e' is first char or char before 'i' is not 'c'.

In the second case we have to check index to prevent checking outside of scope. 

In the first case we don't have to worry about scope since the if condition fails as soon as [i + x] finds null terminator. Of course the input always has to be a string (difference between string and char array is that string is null terminated), otherwise the loop wouldn't even know when to stop.

For loop could also check with condition str[i + 1] since the loop doesn't have to go until the end to know that those two cases can't happen, but the result is the same.

### DIFF
--- a/chall08/rjaakonm.c
+++ b/chall08/rjaakonm.c
@@ -3,9 +3,9 @@ int ft_ie_except_after_c(char *str)
 	for (int i = 0; str[i]; i++)
 	{
 		if (str[i] == 'c' && str[i + 1] == 'i' && str[i + 2] == 'e')
-				return (0);
+			return (0);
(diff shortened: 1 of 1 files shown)


T7Q:
Very short and neat code with one loop only. I liked the simplicity of your solution and description, very well done!

T7Q (approved):
Very short and neat code with one loop only. I liked the simplicity of your solution and description, very well done!

PR #223, opened 2020-05-04
Well yes, but actually no

RandommUser:
English grammar rules don't make sense, but still better than the pronunciations.

**Take 2:** Reversed the string parse direction. Added a checking case for the "ei" when the string starts with it.
Optimized counter calculation to only happen if needed.

I don't like the hardcoded solution for the "ei" case but it will do for now

---

**Take 1:** So, just a simple string looper. Get the string len, have a counter to keep track of positions, sub-function that checks if previous letter was 'c' and next is the third part  of the combo. Inverse the result for the "must not" rule. Keep the loop going as long as the rule hasn't been broken or until there is no possibility to fit the "ie" or "ei" in it.

Default to return to 1 and only change it if the the rule is broken. Could have made it into all one function but somewhat tried to follow the Norm.

### DIFF
--- a/chall08/phakakos.c
+++ b/chall08/phakakos.c
@@ -1,0 +1,36 @@
+long	ft_strlen(char *str)
+{
+	long	i;
+	
+	i = 0;
+	while (str[i])
+		i++;
+	return (i)
+}
+
+int		following(char *str, char c)
+{
+	if (str[0] == 'c' && str[2] == c)
+		return (1);
+	return (0);
+}
+
+int		ft_ie_except_after_c(char *str)
+{
+	int		rtn;
+	long	len;
+	long	i;
+	
+	i = 0;
+	len = ft_strlen(str) - 1;
+	rtn = 1;
+	while (i < len && rtn == 1)
+	{
+		if (str[i] == 'e' && i > 0)
+			rtn = following(&str[i - 1], 'i');
+		else if (str[i] == 'i' && i > 0)
+			rtn = !following(&str[i - 1], 'e')
+		i++;
+	}
+	return (rtn)
+}
--- a/chall08/phakakos.c
+++ b/chall08/phakakos.c
@@ -1,10 +1,14 @@
-long	ft_strlen(char *str)
+long	ft_strlen(char **str)
 {
 	long	i;
 	
 	i = 0;
-	while (str[i])
+	while (str[0][i])
+	{
+		if (str[0][i] >= 'A' && str[0][i] <= 'Z')
+			str[0][i] += 32; 
 		i++;
+	}
 	return (i)
 }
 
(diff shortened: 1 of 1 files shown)


sluhtala (approved):
I think I now understand how does it works. And it seems like it works, so no reason not to approve. I like the fact you considered the upper/lower case words.

chall08/phakakos.c
```diff
@@ -0,0 +1,43 @@
...
+
+int		ft_ie_except_after_c(char *str)
+{
+	int		rtn;
+	long	i;
+
+	rtn = 1;
+	i = 0;
+	if (rtn)
+		i = ft_strlen(&str) - 1;
```
sluhtala, line 30: I wonder why there is "if (rtn)" right after it has been set to 1. It will always be true.

chall08/phakakos.c
```diff
@@ -0,0 +1,43 @@
...
+	long	i;
+	
+	i = 0;
+	while (str[0][i])
+	{
+		if (str[0][i] >= 'A' && str[0][i] <= 'Z')
+			str[0][i] += 32; 
+		i++;
+	}
+	return (i)
```
sluhtala, line 12: I think this changes the original values of str. I try to avoid changing them if the function  doesn't require that. Also doesn't this work with single pointer?

chall08/phakakos.c
```diff
@@ -0,0 +1,43 @@
...
+{
+	if (str[0] == 'c' && str[2] == c)
+		return (1);
+	return (0);
+}
+
+int		ft_ie_except_after_c(char *str)
+{
+	int		rtn;
+	long	i;
```
sluhtala, line 25: Using long seems a bit overkill for ft_strlen. Also using unsigned int (size_t) might be better.

chall08/phakakos.c
```diff
@@ -0,0 +1,43 @@
...
+	i = 0;
+	if (rtn)
+		i = ft_strlen(&str) - 1;
+	while (rtn == 1 && --i >= 0)
+	{
+		if (str[i] == 'e')
+		{
+			if (i == 0 && str[i + 1] == 'i')
+				return (0);
+			rtn = following(&str[i - 1], 'i');
```
sluhtala, line 37: I believe you can pass "&str[i - 1]" as "str + i - 1". But both works.

sluhtala (approved):
Everything seems to work. The problem can be solved much more simply but it looked like you were exploring. Good job!

chall08/phakakos.c
RandommUser: Ah shiet, it was supposed to be str. Damn last minute changes.

chall08/phakakos.c
RandommUser: When you are passed a single pointer you cannot change the original value so it doesn't change the one given to the whole function, but I can change it for my program to only read with lowercase letters.

jgengo:
@sluhtala do we merge it or not?

sluhtala:
Yes, so weird stuff but everything works. Can be merged.

PR #224, opened 2020-05-04
rpehkone 08

vossr:
Iterates string. If "ei" appears in the string, it must immediately follow "c". If "ie" appears in a word, it must not immediately follow "c".  If either of the conditions is not met a zero is returned.


levonml:
I like how compact your code is.  Especially the last update is fantastic! I am proud that I could understand it ;)

jgengo:
@levonml do we merge this PR ?

PR #225, opened 2020-05-04
chall08/vkuikka.c

vkuikka:
The function iterates through the string checking each character for the character 'i' before 'e' and 'e'  before 'i'. If "ei" is after a 'c' then the string follows the rule. Otherwise return zero. The opposite comparison is done if "ie" is found from the string.


mkarkaus (approved):
Nicely compact code in result of not using indexes and I didn't come up with any 'edge case' scenario in which this wouldn't work. Nice job Ville!

PR #226, opened 2020-05-04
Add my solution to chall08

hiiri:
Iterate over char array. If 'e' or 'i' is found, send string and index to equivalent function (check_ei or check_ie).
"check_ei" checks if there's "ei", if yes, it checks for a 'c' before it.
"check_ie" checks if there's "ie", if yes, it checks that there is no 'c' before it.
If any of these return 0, the string does not obey the rule, so return 0.
Otherwise if nothing returned 0, the string does obey it. Return 1 in that case.

### DIFF
--- a/chall08/alcohen.c
+++ b/chall08/alcohen.c
@@ -1,0 +1,50 @@
+#include <stdlib.h>
+
+/*
+**  Return 1 if "ei" after "c", 0 otherwise
+*/
+static int check_ei(char *str, size_t i)
+{
+    if (str[i + 1] && str[i + 1] == 'i')
+    {
+        if (str[i - 1] && str[i - 1] == 'c')
+            return (1);
+        else
+            return (0);
+    }
+    return (1);
+}
+
+/*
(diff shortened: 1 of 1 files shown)


chall08/alcohen.c
```diff
@@ -0,0 +1,50 @@
+#include <stdlib.h>
+
+/*
+**  Return 1 if "ei" after "c", 0 otherwise
+*/
+static int check_ei(char *str, size_t i)
+{
+    if (str[i + 1] && str[i + 1] == 'i')
+    {
+        if (i > 0 && str[i - 1] && str[i - 1] == 'c')
```
saraarmiida, line 10: Is checking str[i - 1] necessary after already checking if i > 0? Also function could be made shorter with checking for the condition where it returns 0 first ( i = 0 or str[i - 1] != 'c'). But works perfectly fine like that :)

saraarmiida (approved):
Code could probably be made shorter and without helper functions but it seems to work perfectly fine in most cases. Nice code🙌

PR #227, opened 2020-05-04
Add chall08 mkarkaus

mkarkaus:
I created a function that scrolls through the input string using index and in case the character is either 'i' or 'e' function checks the characters next to it and if they don't follow the rule function returns 0. Otherwise it returns 1.

### DIFF
--- a/chall08/mkarkaus.c
+++ b/chall08/mkarkaus.c
@@ -1,0 +1,15 @@
+int		ft_ie_except_after_c(char *str)
+{
+	int		i;
+
+	i = 0;
+	while (str[i] != '\0')
+	{
+		if (str[i] == 'e' && str[i + 1] == 'i' && str[i - 1] != 'c')
+			return (0);
+		if (str[i] == 'i' && str[i + 1] == 'e' && str[i - 1] == 'c')
+			return (0);
+		i++;
+	}
+	return (1);
+}


chall08/mkarkaus.c
```diff
@@ -0,0 +1,15 @@
+int		ft_ie_except_after_c(char *str)
+{
+	int		i;
+
+	i = 0;
+	while (str[i] != '\0')
```
salahadawi, line 6: Can also be shortened to just `while (str[i])`

chall08/mkarkaus.c
```diff
@@ -0,0 +1,15 @@
...
+	{
+		if (str[i] == 'e' && str[i + 1] == 'i' && str[i - 1] != 'c')
+			return (0);
+		if (str[i] == 'i' && str[i + 1] == 'e' && str[i - 1] == 'c')
+			return (0);
```
salahadawi, line 11: With a string such as "ei", `str[i - 1]` tries to access the character at index -1, which is out of bounds. This results in undefined behavior and might cause a segfault. You should check in some way that you aren't accessing the string from an invalid index.

salahadawi (approved):
Hello Miikka! Your function is nicely written, it's clean and simple. A small edge case might cause some problems, but otherwise it seems to work great. Good work!

chall08/mkarkaus.c
mkarkaus: Hey Sala! Had to test it out with a main.c and strangely enough strings "ei" and "ie" don't cause any seg faults or anything of that sort. I Guess it seg faults only if you're trying to print the str[i - 1] or replace it with something but not if you only compare it to something.

salahadawi:
```
int main()
{
    char *str = "cie";
    char *str2 = &str[1];
    
    printf("%d\n", ft_ie_except_after_c("ie"));
    printf("%d\n", ft_ie_except_after_c(str2));
}
```
Testing it like this, str2 is equal to "ie", and contains the character 'c' at index -1. Even though the strings are the same in both printf functions, one of them prints 0 and the other rprints 1. This is because index -1 is not a part of str2, but it is still checked and here it happens to contain the character 'c'.

mkarkaus: Those values do work as you said, interesting... That case didn't even cross my mind. Well spotted!

salahadawi: Yeah, that index -1 could contain anything, and if it happens to contain for example data that you aren't allowed to access, that would result in a segfault. Not exactly sure how to test that though.

PR #228, opened 2020-05-04
placeholder

tuommii:
Looking for "ie" or "ei" strings with `strcasestr` (that's why i have `#define _GNU_SOURCE`)
and then checking if "c" is found right before them. Repeating (not optimized, increasing str instead of returned pointer) until whole string is checked.

### DIFF
--- a/chall08/mtuomine.c
+++ b/chall08/mtuomine.c
@@ -2,35 +2,19 @@
 #include <string.h>
 
 static int c_found(char *str) {
-	if (!str || !(str-1))
-		return (0);
 	char c = *(str - 1);
 	return (c && (c == 'c' || c == 'C'));
 }
 
 int ft_ie_except_after_c(char *str) {
-	if (!str)
-		return (1);
-
 	char *ei_found = strcasestr(str, "ei");
 	char *ie_found = strcasestr(str, "ie");
 
-	if (ei_found == 0 && ie_found == 0) {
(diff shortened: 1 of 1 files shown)


chall08/mtuomine.c
```diff
@@ -0,0 +1,18 @@
...
+}
+
+int ft_ie_except_after_c(char *str) {
+	char *ei_found = strcasestr(str, "ei");
+	char *ie_found = strcasestr(str, "ie");
+
+	if ((ei_found && !c_found(ei_found)) || (ie_found && c_found(ie_found)))
+		return (0);
+	if ((ei_found && c_found(ei_found)) || (ie_found && !c_found(ie_found)))
+		return ft_ie_except_after_c(str+=2);
```
vkuikka, line 16: As you said (if I understood correctly) you could be using *ei_found and *ie_found pointers here instead of *str so strcasestr doesn't have to check the same stuff again if the match is far from the beginning of the string.

vkuikka (approved):
Recursion and no loops makes the code really easy to read. Code looks very nice. Good job!

jgengo:
@vkuikka do we accept this merge?

vkuikka:
> @vkuikka do we accept this merge?

Yes. I thought approving changes means it can be merged so what should I have done? I'm just wondering why you ask some people this.

PR #229, opened 2020-05-04
tkuumola solution chall08

T7Q:
1.  first input validation, if there is a string.
2.  then created a match function that checks with strstr from sting.h library if there are any occurrences of the needle in the string (1 if found, 0 if not)
3. using match function, identify the words that match the rule: 1) without "ei" and "ie", 2) with "cei" and not "cie" 3) with "ie" and not "cie" and not "ei" (to eliminate word like deifies).

### DIFF
--- a/chall08/tkuumola.c
+++ b/chall08/tkuumola.c
@@ -1,12 +1,33 @@
 #include <string.h>
 
+int match(char *str, char *needle)
+{
+    int flag = 0;
+
+    while( (str=strstr(str, needle)) != NULL)
+    {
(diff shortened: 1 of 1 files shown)


lkallio (changes requested):
EDIT: realized that this was supposed to be done tomorrow, disregard.

T7Q:
Lalli, thank you for the early review!!! seems that I misunderstood the task, made corrections based on your suggestions. Come back tomorrow to check again :)

PR #230, opened 2020-05-04
Why not

lkallio:
Why not.


chall08/lkallio.c
```diff
@@ -0,0 +1,7 @@
+int		ft_ie_except_after_c(char *str)
+{
+	return (!str || !str[0] || !str[1] || !str[2]
+	|| (!(str[1] == 'i' && str[2] == 'e' && str[0] == 'c')
+	&& !(str[1] == 'e' && str[2] == 'i' && str[0] != 'c')
```
kurval, line 5: This might be a problem if a  word start with 'ei'

kurval (approved):
This was quite smart solution to use recursion but there is one special case I was thinking. What if a word starts with 'ei' or that is a single word? I would still approve this pull request because I'm not sure does it need to handle those words?

PR #231, opened 2020-05-04
chall08 (jnovotny)

jiricodes:
**Solution**
Firstly we check for starting conditions:
- if given string doesn't exist -> return 1
- if given string starts with "ei" -> return 0
- if string is shorter than 3 letters or doesn't contain substrings in question ("ei", "ie") -> return 1

Any present string at this point has to be 3 or more characters long and include at least one occurrence of "ei" or "ie". This allows us to safely iterate over the string starting from second position till the second last position and checking for out rules conditions. If we find any violation of given rules, we return 0. Otherwise upon reaching the end return 1.

**Reasoning**
The reason to eliminate certain words at the beginning is twofold. Firstly there's only 1 word shorter than 3 characters that violates given rules - "ei", so it's faster to check for this special case. Plus it allows to iterate over the string using  `i - 1, i, i + 1 starting from i = 1` without further constrains or danger of segfault. Secondary in general standard library functions such as strcmp and strstr are well optimised and faster than iteration over the string manually, so we should return results much faster in case of "ei" and "ie" not present in the string without hindering performance too much otherwise.


chall08/jnovotny.c
```diff
@@ -0,0 +1,36 @@
...
+
+int ft_ie_except_after_c(char *str)
+{
+	int i;
+
+	if (!str)
+		return 1;
+	else if (strlen(str) >= 2 && str[0] == 'e' && str[1] == 'i')
+		return 0;
+	else if (strlen(str) < 3 || (!strstr(str, "ei") && !strstr(str, "ie")))
```
hakolao, line 23: :+1: nice to save calculations with these

hakolao (approved):
Good job, seems to work fine.

Simple solution, didn't really have anything to comment about this :)

PR #232, opened 2020-05-04
chall08 by sluhtala

sluhtala:
Hello,

- if null str, return -1. If the word is "" empty, return 0.
- if the word starts with "ei".. it means there is no "c" in the beginning
- looping all the letters in the word and checking if the two rules don't break.
- if rule is not true, return 0 and if word is all checked, return 1.

I used this function with get_next_line and checked the file /usr/share/dict/web2 (235886 English words) and found 2765 words that doesn't follow the rule. ~1% ... interesting :D

### DIFF
--- a/chall08/sluhtala.c
+++ b/chall08/sluhtala.c
@@ -1,0 +1,27 @@
+#include <string.h>
+
+int ft_ie_except_after_c(char *str)
+{
+	int i;
+
+	i = 0;
+	if (str == NULL)
+		return (-1);
+	if (strlen(str) == 0)
+		return (1);
+	if (str[0] == 'e' && str[1] == 'i')
+		return (0);
+	while (str[i])
+	{
+		if (str[i] != 'c' && str[i + 1] == 'e' && str[i + 2] == 'i')
+		{
+			return (0);				
+		}
+		if (str[i] == 'c' && str[i + 1] == 'i' && str[i + 2] == 'e')
+		{
+			return (0);
+		}
+		i++;
+	} 
+	return (1);
+}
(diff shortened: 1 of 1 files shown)


RandommUser:
_Review in progress_

chall08/sluhtala.c
```diff
@@ -0,0 +1,25 @@
+int ft_ie_except_after_c(char *str)
+{
+	int i;
+
+	i = 0;
+	if (str == NULL)
+		return (-1);
```
RandommUser, line 7: Good to check the string exists. You could go all philosophical about the return values since NULL and "\0" both basically have a value 0... Buut also I don't see an empty string being a word to begin with so just having checks that prevent the program from crashing is good.

chall08/sluhtala.c
```diff
@@ -0,0 +1,25 @@
...
+	while (str[i])
+	{
+		if (str[i] != 'c' && str[i + 1] == 'e' && str[i + 2] == 'i')
+		{
+			return (0);				
+		}
+		if (str[i] == 'c' && str[i + 1] == 'i' && str[i + 2] == 'e')
+		{
+			return (0);
+		}
```
RandommUser, line 21: A good way to check the letter rule. Except you don't check if you are going over the array, so that will cause some crashes with some cases.

You also previously didn't have brackets with a single line if-blocks so having them here is a bit strange. Try to stick to one format.

RandommUser (approved):
Simple solution for a simple problem. You could have probably made the while-loop handle the second and third pre-loop if statements, but should still work on most cases. Approved.

chall08/sluhtala.c
rasmusjaa: Just commenting that if statement won't go over the array, because it stops if (str[i + 1]) finds null terminator so it won't check (str[i + 2]). If there is no null terminator while loops wouldn't stop anyway so we have to assume that only null terminated strings are given to the function, otherwise we would need to receive the length of the char array too.

PR #233, opened 2020-05-04
chall08 by spentti

saraarmiida:
I index through the string given as a parameter and return 0 with two conditions:
1. if there is  'c' followed by 'i' and 'e'
2. if there is 'e' followed by 'i' and the 'e' is either the first character of the string or there is no 'c' before it
if neither of these conditions are met, just return 1, the string has passed the test!

### DIFF
--- a/chall08/spentti.c
+++ b/chall08/spentti.c
@@ -1,0 +1,15 @@
+int ft_ie_except_after_c(char *str)
+{
+	int i;
+
+	i = 0;
+	while (str[i + 1] != '\0')
+	{
+		if (str[i] == 'c' && str[i + 1] == 'i' && str[i + 2] == 'e')
+			return (0);
+		if ((i == 0 || str[i - 1] != 'c') && str[i] == 'e' && str[i + 1] == 'i')
+			return (0);
+		i++;
+	}
+	return (1);
+}


chall08/spentti.c
```diff
@@ -0,0 +1,15 @@
+int ft_ie_except_after_c(char *str)
+{
+	int i;
+
+	i = 0;
+	while (str[i + 1] != '\0')
```
rasmusjaa, line 6: Nice condition, doesn't check anything it doesn't have to.

chall08/spentti.c
```diff
@@ -0,0 +1,15 @@
...
+	{
+		if (str[i] == 'c' && str[i + 1] == 'i' && str[i + 2] == 'e')
+			return (0);
+		if ((i == 0 || str[i - 1] != 'c') && str[i] == 'e' && str[i + 1] == 'i')
+			return (0);
```
rasmusjaa, line 11: Seems to have all necessary checks and nothing else, very good. One small optimization might be to check str[i] == 'e' first since (i == 0 || str[i - 1] != 'c') is true in a lot more cases so function makes more checks. Really small difference, but everything else is optimized nicely so thought I would point this out even if there's no difference for end result :)

rasmusjaa (approved):
Very good code, probably the most effective solution since it loops through only once and checks only necessary things. Clean and nice compact code!

PR #235, opened 2020-05-05
v1.0 ready

MikeyLHolm:
Checking current ,next and previous index to find a words that don't use the rule. Otherwise return 1 at end.


lkallio (changes requested):
If you use str[i - 1], you can't start looping from 0, that will make it search for -1 index of the string.

MikeyLHolm:
> If you use str[i - 1], you can't start looping from 0, that will make it search for -1 index of the string.

That is true, should have made that different.

jgengo:
@lkallio do we merge it or not ?

PR #236, opened 2020-05-05
Adding chall08

kurval:
Solution to chall08. My solution was to check if there was 'e' or 'i' letter inside of given word and check the letter after (I used len - iterator to check if  there is a letter left). In case of 'ei' I also had to check that 'c' comes before that, so iterator had to be greater that 0. If that was the case I returned  1. Same applies to 'ie' but now I had to check that letter before was different than 'c' except if 'i' and 'e' are the first two letters (I'm not sure if those word exist anyway?). 

If none of them wasn't true BUT 'ie' or 'ei' exist in a word I assigned value exc to 0. If none of them ('ie' or 'ei') appears in a word I can return exp 1 because it follows the rule.


chall08/vkurkela.c
```diff
@@ -0,0 +1,40 @@
...
+    
+    i = 0;
+    len = 0;
+    exc = 1;
+    while (str[i++] != '\0')
+        len++;
+    i = 0;
+    while (str[i] != '\0')
+    {
+        if (str[i] == 'e' && (len - (i + 1)) >= 1)
```
lapaset, line 15: The condition would be more clear as i + 1 < len or i < len - 1

chall08/vkurkela.c
```diff
@@ -0,0 +1,40 @@
...
+        {
+            if (i != 0)
+            {
+                if (str[i - 1] == 'c' && str[i + 1] == 'i')
+                    return (1);
```
lapaset, line 20: Returning at this point means that the rest of the string will not be validated. Instead you could return 0 when the condition doesn't fill.

chall08/vkurkela.c
```diff
@@ -0,0 +1,40 @@
...
+            }
+            if (str[i + 1] == 'i')
+                exc = 0;
+        }
+        else if (str[i] == 'i' && (len - (i + 1)) >= 1)
+        {
+            if (i != 0)
+            {
+                if (str[i - 1] != 'c' && str[i + 1] == 'e')
+                    return (1);
```
lapaset, line 30: Same as above

chall08/vkurkela.c
```diff
@@ -0,0 +1,40 @@
...
+            }
+            else if (i == 0 && str[i + 1] == 'e')
+                return (1);
```
lapaset, line 33: Same as above

lapaset (approved):
While the logic was there, there were some errors that i think would lead to failing tests.

As you assign the exc flag to 1 at the beginning, it would be enough to just change it to 0 when you encounter a failing letter combination. Or instead of returning the 1 when true you could return 0 when the condition doesn't fill.

There are no excess files and the file goes by the rules, so I would accept this pr :)

chall08/vkurkela.c
kurval: Thanks Liisa for the feedback! :) Yeah so my idea for here was to return 1 immediately because when this condition fills I am sure it follows the rule and I don't need to validate rest of the string. Same applies to other cases where I'm return 1 straight away. So I was wondering is there some reason why would you validate rest of the string?

lapaset: If the string has more than one combination of ei or ie you need to check if the rest of them follow the rule as well to confirm that the whole string follows the rule. Returning 0 at the first instance would be ok, since you know that the string does not follow the rule.

chall08/vkurkela.c
kurval: This would be more clear definitely. ^^for the upper comment the point why I assingned to 1 at the begin was that if none of 'ie' or 'ei' appear in a word. If either one appears in a word I'm assigning it to 0. I know this isn't the most delicate solution

chall08/vkurkela.c
kurval: that is true! I wasn't thinking clear enough. I should have return 0 if condition doesn't fill

PR #237, opened 2020-05-05
Add ohakolas solution

hakolao:
If string is `NULL` return `1` (could also be 0, up to creator in my opinion) if `str == "ei"` return 0, else:
1. Test case `ei` for `str`
2. Test case `ie` for `str`

Tests for `ie` and `ei` find their corresponding substring using `strstr` from copied `str` pointer `s`. If they encounter that sub string (`s - str > 0`), check the previous character at that position if it's what it's supposed to be. If not, return false.

`s++;` ensures that while loop reaches NULL eventually.

This code was tested with (not so thorough):
```
#include <stdio.h>

int   main(void)
{
  printf(
    "test1: %d\n"
    "test2: %d\n"
    "test3: %d\n"
    "test4: %d\n"
    "test5: %d\n",
    ft_ie_except_after_c("a") == 1,
    ft_ie_except_after_c("zombie") == 1,
    ft_ie_except_after_c("transceiver") == 1,
    ft_ie_except_after_c("veil") == 0,
    ft_ie_except_after_c("icier") == 0);
  printf("test6: %d\n", ft_ie_except_after_c("hierarchy") == 1);
  printf("test7: %d\n", ft_ie_except_after_c("receiptcie") == 0);
  printf("test8: %d\n", ft_ie_except_after_c("receipt") == 1);
  printf("test9: %d\n", ft_ie_except_after_c("aaaaaaaaaaaaaa") == 1);
  printf("test10: %d\n", ft_ie_except_after_c("eieieieieie") == 0);
  printf("test11: %d\n", ft_ie_except_after_c("ceiceiceiceiceie") == 1);
  printf("test12: %d\n", ft_ie_except_after_c(NULL) == 1);
  printf("test13: %d\n", ft_ie_except_after_c("") == 1);
  printf("test14: %d\n", ft_ie_except_after_c("ie") == 1);
  printf("test15: %d\n", ft_ie_except_after_c("ei") == 0);
}
```

### DIFF
--- a/chall08/ohakola.c
+++ b/chall08/ohakola.c
@@ -1,18 +1,15 @@
 #include <string.h>
 #include <stdbool.h>
 
-static int	test(char *str, char *test, bool must_follow_c)
+static int	test(char *str, char *test, bool follows_c)
 {
-	char	*strcpy;
-
-	strcpy = str;
-	while ((strcpy = strstr(strcpy, test)))
+	char	*s = str;
+	while ((s = strstr(s, test)))
 	{
-		if (strcpy - str > 0)
-			if (must_follow_c ? str[strcpy - str - 1] != 'c' :
-				str[strcpy - str - 1] == 'c')
+		if (s - str > 0)
+			if (follows_c ? str[s - str - 1] != 'c' : str[s - str - 1] == 'c')
 				return (false);
-		strcpy++;
+		s++;
 	}
 	return (true);
 }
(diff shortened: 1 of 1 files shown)


jiricodes (approved):
Nice to see a solution using strstr(). I thought myself it might be neat and it definitely looks that way in your submission. The ternary gave me a scare at first, but it made the whole if statement a bit cleaner in the end. Cool, clean, readable and well described working solution - Merge!

PR #239, opened 2020-05-05
Chall08

tangxing1220:
English isn't my mother language. I didn't know this English spelling rule. Firstly, I study the rule. 
I find out two things are critical to chall08.  
 
To Assume of all words following the rule: 
Based on the examples, I assume all words follow the rule (return=1), the task is finding the exception(return=0). As while I don't care about big letter and the case of word has e letter twice.

The design of testing cases:
1. ""  " "   "a"   "ab"  "abc"   " abcfg"   "xxixxxx"    "xxexx" ,    "xxxxxe",    "xxxxi"  "exxxx" "ixxxx" ....
2. "i"   "e"   "ixxx"  "exxx",    "iexx"  "eixx"  "ciexx"   "ceixx" .... 
3. "xxxxceixxx",   "xxxxciexx"   ...
4, "xxxxxiexx',      "xxxxxeixxx"  ...
5, "xxxxxxce"     "xxxxxxci"     "xxxcie"     " xxxcei"  ...
6,  "xxxei"   "xxxie"   ....

using while loop to go through every char of the string. and checking beside chars by array indexs.


chall08/xtang.c
```diff
@@ -0,0 +1,42 @@
...
+		if (str[i] == 'e')
+			break ;
+		i++;
+	}
+	i = 0;
+	while (str[i] != '\0')
+	{
+		if (str[i] == 'e')
+		{
+			if (str[i - 2] == 'c' && str[i - 1] == 'i')
```
hiiri: If i is 0 or 1, this will try to go to memory before str[0] which is something it shouldn't access (str[-1] or str[-2]).

chall08/xtang.c
```diff
@@ -0,0 +1,42 @@
...
+				return (0);
+			if (str[i - 1] == 'c' && str[i + 1] == 'i')
+				return (1);
+			if (str[i - 1] == 'i' && str[i - 1] != 'c')
+				return (1);
+			return (1);
```
hiiri, line 43: This returns 1 on words like "veil". There needs to be a check something like this: 
`if (str[i - 1] != 'c' && str[i + 1] == 'i')
				return (0);`

hiiri (approved):
Cool. Some ways this could be improved:
1. Separate function(s) for checks. For example something like 
`int checker(char *str, char *chars_to_check, char before_chars_to_check);`
Then it could work like 
```
main()
{
     return (checker(str, "ei", 'c'))
}
```
And that would check if there's "ei" and a 'c' before it for example. If you had to change your code to add/modify rules, it is helpful to have modularity using functions.

chall08/xtang.c
```diff
@@ -0,0 +1,42 @@
...
+		if (str[i] == '\0')
+			return (1);
+		if (str[i] == 'e')
+			break ;
+		i++;
+	}
+	i = 0;
+	while (str[i] != '\0')
+	{
+		if (str[i] == 'e')
```
hiiri, line 20: I like how you check == 'e' and use that to check for "ie" and "ei" at the same time. Clever.

chall08/xtang.c
tangxing1220: To update my code. There is a more clear logical judgement now.

